PR #43, opened 2018-05-18
Special case for openid auth scheme

theferrit32:


trel:
Need to squash this down to a single commit.

And reference irods/irods#3945 in the commit message.

PR #46, opened 2018-05-23
[irods/irods#3945] add provider to iinit login

theferrit32:
Without this, an execution of iinit will pick up the existing session from the user's system (if it exists from a prior iinit) and not perform a full re-authentication flow.  Specifying a context string to clientLoginOpenID overrides that behavior in the auth plugin.


trel:
running through CI

trel:
passed CI

Issue #47, opened 2018-05-24
iadmin can only change passwords when connected to a catalog provider server


tempoz:
Currently, `iadmin moduser` makes use of the challenge string provided by the server and the `ObfEncodeByKeyV2` function to encode passwords. This challenge string is generated on a per-server basis, so when the password request change is initiated  forwarded from a resource server to a catalog server to actually decode, scramble, and store the new password in the database, the decode fails.

This would be an acceptable limitation if the interface required it. However, the generalAdmin interface passes what it's sent on to the database plugin in the exact same way that the interface `ipasswd` uses (userAdmin) does, and `ipasswd` does not make use of or require a challenge, and, indeed, works when connected to a catalog consumer server.

This is because the database plugin's mod user function will decode passwords WITH OR WITHOUT a challenge encoded in them, based on whether or not the challenge was encoded. Therefore, we can just encode passwords for `iadmin moduser` in the same way we do for `ipasswd`. Because they are still encoded using a known password, they are still secure to the same degree any API call made by `ipasswd` is.

PR #48, opened 2018-05-24
[#47] Change iadmin to use obfEncodeByKey

tempoz:
obfEncodeByKeyV2 requires the challenge string from the server, which confounds decoding when not connected directly to a catalog provider. obfEncodeByKey works solely based on encoding the new password with another password (in all cases currently, specifically the password of the user calling it). This allows us to change the password even if we are not connected to a catalog provider server.


trel:
nearly passed CI (hiccup) - rerunning

trel:
passed CI

PR #56, opened 2018-06-28
Ask for SSL encrypted communication during first iinit call

joergsteinkamp:
Include a question if default SSL parameters in ~/.irods/irods_environment.json should be included. This is required to communicate with a Server with irods_client_server_policy=CS_NEG_REQUIRE.

### DIFF
--- a/src/iinit.cpp
+++ b/src/iinit.cpp
@@ -249,6 +249,59 @@ main( int argc, char **argv ) {
             json_string( my_env.rodsAuthScheme ) );
     }
 
+    // mz_js_20180626+
+    if ( strlen( my_env.rodsClientServerNegotiation ) == 0 ) {
+      if ( !doingEnvFileUpdate ) {
+        doingEnvFileUpdate = true;
+        printUpdateMsg();
+      }
+      printf( "Include default SSL encrypted communication [Y/n]: " );
+      std::string response = "";
+      char inbuf[100];
+      getline( std::cin, response );
+      strncpy( inbuf, response.c_str(), 90 );
+      if ( strncmp( inbuf, "n", 1 ) != 0 || strncmp( inbuf, "N", 1 ) != 0 ) {
+        snprintf(
+		 my_env.rodsClientServerNegotiation,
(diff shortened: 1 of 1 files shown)


src/iinit.cpp
```diff
@@ -249,6 +249,59 @@ main( int argc, char **argv ) {
...
 
+    // mz_js_20180626+
+    if ( strlen( my_env.rodsClientServerNegotiation ) == 0 ) {
+      if ( !doingEnvFileUpdate ) {
+        doingEnvFileUpdate = true;
+        printUpdateMsg();
+      }
+      printf( "Include default SSL encrypted communication [Y/n]: " );
+      std::string response = "";
+      char inbuf[100];
```
trel: magic number?   is the 90 related below?  can we name this length and use as a variable?  or not declare a length at all?

src/iinit.cpp
```diff
@@ -249,6 +249,59 @@ main( int argc, char **argv ) {
...
+      getline( std::cin, response );
+      strncpy( inbuf, response.c_str(), 90 );
+      if ( strncmp( inbuf, "n", 1 ) != 0 || strncmp( inbuf, "N", 1 ) != 0 ) {
```
trel: need to think through whether we want this on by default...

or just make it available at first with default still 'N'?

joergsteinkamp: I copied those two "magic numbers" 100 and 90 from iquest.cpp line 178, where there is also a Y/n question. 

With respect to the default reply: it's safer to use SSL, but more comfortable to not using it. I prefer to force SSL, since it is my task to provide a secure environment for the scientists especially for medical data.

trel:
@alanking please eyeball this as well - something we want to include/incorporate with the 4.3.0 `iinit` work?

korydraughn:
> gotcha - we'll need to create a separate issue, and connect this one with that (in 4.3.1)

Done

trel:
now targeted for https://github.com/irods/irods/issues/6398

alanking:
Superseded by #338.

PR #86, opened 2019-02-12
[irods/irods2082] update iexit to always remove saved credential file

trel:
tests included in https://github.com/irods/irods/pull/4263

### DIFF
--- a/src/iexit.cpp
+++ b/src/iexit.cpp
@@ -11,12 +11,9 @@ main( int argc, char **argv ) {
 
     signal( SIGPIPE, SIG_IGN );
 
-    int status, ix;
     rodsArguments_t myRodsArgs;
-    rodsEnv myEnv;
-    char *envFile;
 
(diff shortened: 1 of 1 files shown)


src/iexit.cpp
```diff
@@ -25,40 +22,37 @@ main( int argc, char **argv ) {
...
-                obfRmPw( 0 );
-            }
-            else {
-                obfRmPw( 1 );
-            }
-        }
-        else {
-            printf( "option %s is unrecognized\n", argv[ix] );
-        }
+    if ( myRodsArgs.verbose == True ) {
```
alanking: Based on `test_iexit_verbose` in irods/irods#4263, is the verbose option supposed to wait for confirmation from the user? I don't see it here, but I could be misunderstanding the intent.

Either way, 40-41 be part of the block which starts on line 34. Or is the unlink supposed to happen after the print on line 35? In the latter case, that print could be moved to the block started on line 28.

trel: Yes, it does wait for confirmation.   I'll look at collapsing the if/else.

PR #109, opened 2020-02-06
[irods#4299] Added spdlog dependency to build script (master)

korydraughn:
Because the icommands include **irods_re_plugin.hpp**, the spdlog dependency must be added to the build script even though icommands don't to use spdlog.

spdlog is an implementation detail meant for server-side use only.

This could possibly be avoided by using macros.  In particular `IRODS_ENABLE_SYSLOG`. See https://github.com/irods/irods/blob/e563c57c6813463aa731fecd04e368da5a23b138/lib/core/include/apiHandler.hpp#L95-L131


korydraughn:
This PR can be closed if we use the following code: https://github.com/irods/irods/pull/4701/commits/6bd50d51ae2ce09b011ce0a418512ff94bb076b3

It uses the `IRODS_ENABLE_SYSLOG` macro to determine whether `irods_logger.hpp` is included or not. It also removes code through the use of these macros.

trel:
seems good - let's close this one

PR #114, opened 2020-02-14
[irods/irods#4506] Add iunreg, deprecate irm -U (4-2-stable)

alanking:
CI tests are running.

Requires irods/irods#4730

### DIFF
--- a/src/iunreg.cpp
+++ b/src/iunreg.cpp
@@ -1,0 +1,156 @@
+#include "rodsClient.h"
+#include "parseCommandLine.h"
+#include "rodsPath.h"
+#include "trimUtil.h"
+#include "rmUtil.h"
+#include "irods_client_api_table.hpp"
+#include "irods_pack_table.hpp"
+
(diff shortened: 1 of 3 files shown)


src/iunreg.cpp
```diff
@@ -0,0 +1,162 @@
+/*** Copyright (c), The Regents of the University of California            ***
```
trel: not california

src/iunreg.cpp
```diff
@@ -0,0 +1,162 @@
...
+ *** For more information please refer to files in the COPYRIGHT directory ***/
+/*
+ * iunreg - The irods replica trimming utility
```
trel: not replica trimming

trel:
merged to master - so 4-2 can get merged if you mark as ready with a #

alanking:
#s added. Alleviates some of the test failures, but log flushing continues to be an ongoing issue.

PR #115, opened 2020-03-02
[irods/irods#4740] Add iadmin modrepl and ls (master)

alanking:
iadmin modrepl can be used to modify a number of different columns in
R_DATA_MAIN. Some have been restricted until such a time as a valid use
case justifies its exposure in the client. Until then, only a subset of
the columns are modifiable via iadmin modrepl.

iadmin ls now lists the GenQuery attributes and their values which map to
the columns of R_DATA_MAIN. This is mainly for convenience.

---

CI tests running

### DIFF
--- a/src/iadmin.cpp
+++ b/src/iadmin.cpp
@@ -1,21 +1,22 @@
-/*** Copyright (c), The Regents of the University of California            ***
- *** For more information please refer to files in the COPYRIGHT directory ***/
-
-/*
-   Initial version of an administrator interface
-*/
-
-#include "rodsClient.h"
-#include "parseCommandLine.h"
-#include "irods_string_tokenize.hpp"
+#include "filesystem.hpp"
 #include "irods_client_api_table.hpp"
 #include "irods_pack_table.hpp"
+#include "irods_string_tokenize.hpp"
+#include "key_value_proxy.hpp"
+#include "parseCommandLine.h"
+#include "query_builder.hpp"
+#include "rodsClient.h"
 
-#include <iostream>
 #include <algorithm>
+#include <iostream>
+#include <string>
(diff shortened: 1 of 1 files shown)


alanking:
CI tests passed.

src/iadmin.cpp
```diff
@@ -1621,6 +1831,7 @@ void usageMain() {
         " rum (remove unused metadata (user-defined AVUs)",
         " asq 'SQL query' [Alias] (add specific query)",
         " rsq 'SQL query' or Alias (remove specific query)",
+        " modrepl logical_path [ replica_number <int> | resource <string> ] ATTR_NAME VALUE",
```
trel: also modrepl data_id...

src/iadmin.cpp
```diff
@@ -38,6 +39,269 @@ int lastCommandStatus = 0;
...
+    else if ("resource_hierarchy" == option) {
+        v = input.data();
+    }
+    return v;
+} // get_replica_value
+
+auto modify_replica(
+    char** tokens) -> int
+{
+    //const decltype(genquery_attrs)::key_type genquery_attrs_blacklist = {
```
trel: remove comment

src/iadmin.cpp
```diff
@@ -1582,11 +1793,10 @@ void usageMain() {
         " luan Name (list users associated with auth name (GSI/Kerberos)",
         " lt [name] [subname] (list token info)",
         " lr [name] (list resource info)",
-        " ls [name] (list directory: subdirs and files)",
+        " ls logical_path [ replica_number <int> | resource <string> ] (list replica info)",
```
trel: also ls data_id

PR #117, opened 2020-03-03
[irods/irods#4740] Add iadmin modrepl and ls (4-2-stable)

alanking:
iadmin modrepl can be used to modify a number of different columns in
R_DATA_MAIN. Some have been restricted until such a time as a valid use
case justifies its exposure in the client. Until then, only a subset of
the columns are modifiable via iadmin modrepl.

iadmin ls now lists the GenQuery attributes and their values which map to
the columns of R_DATA_MAIN. This is mainly for convenience.

---

CI tests running

### DIFF
--- a/src/iadmin.cpp
+++ b/src/iadmin.cpp
@@ -1,21 +1,22 @@
-/*** Copyright (c), The Regents of the University of California            ***
- *** For more information please refer to files in the COPYRIGHT directory ***/
-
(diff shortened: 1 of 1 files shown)


alanking:
CI tests passed.

trel:
looks good - please #

alanking:
#'d

PR #119, opened 2020-03-07
[#4559][#4682] imeta now handles relative paths and trailing slashes (master)

korydraughn:


korydraughn:
Tests passed.  Added #

PR #129, opened 2020-04-27
Updated help text (4-2-stable)

korydraughn:


trel:
seems good - please point to irods/irods issue numbers

PR #138, opened 2020-07-17
[irods/irods#5000] istream supports resource names, replica numbers, and computing checksums (master)

korydraughn:
This depends on irods/irods#5038

### DIFF
--- a/src/istream.cpp
+++ b/src/istream.cpp
@@ -1,6 +1,4 @@
 #include "rodsClient.h"
-#include "irods_client_api_table.hpp"
-#include "irods_pack_table.hpp"
 #include "connection_pool.hpp"
 #include "dstream.hpp"
 #include "transport/default_transport.hpp"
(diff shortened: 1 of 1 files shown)


src/istream.cpp
```diff
@@ -35,20 +36,17 @@ auto write_data_object(rodsEnv& env, const po::variables_map& vm, io::client::de
...
     desc.add_options()
-        ("help,h", "Produce this message.")
-        ("offset,o", po::value<int_type>()->default_value(0), "The position of the read/write pointer.")
-        ("count,c", po::value<int_type>()->default_value(all_bytes), "The number of bytes to read/write.")
-        ("no-trunc", po::bool_switch(), "Do not truncate the data object on write.")
-        ("append,a", po::bool_switch(), "Append to the data object on write.")
-        ("stream_operation", po::value<std::string>(), "The stream operation (e.g. read or write).")
-        ("logical_path", po::value<std::string>(), "The logical path of a data object.");
+        ("help,h", "")
+        ("R,R", po::value<std::string>(), "")
```
trel, line 42: R,R ?   resource,R instead?

korydraughn: Boost program options does not let you specify only short names. To get around that, you have to give it two short names. Essentially, this means the program treats -R and --R the same.

There may be a hook for controlling this behavior better, but this was the solution I found online.

korydraughn: The --R is not documented in the help text either.

Thoughts?

trel: what you've got below seems fine...

`The root resource to read from or write to.`

trel: happy to have a `--resource,-R` as long/short option... always good to nudge people to use all the letters.

korydraughn: Ok. I'll add it. It makes the code more readable too.

korydraughn:
Added #

PR #139, opened 2020-08-08
[irods/irods#5000] istream supports resource names, replica numbers, and computing checksums (4-2-stable)

korydraughn:
Requires irods/irods#5041


trel:
# if passing

korydraughn:
Added #

PR #150, opened 2020-09-17
New icommand: itouch (master)

korydraughn:


src/itouch.cpp
```diff
@@ -0,0 +1,399 @@
...
+    // Create a new data object if the path does not end with a path separator.
+    io::client::default_transport tp{_conn};
+
+    // Create the data object at the leaf resource if the user provided one.
+    if (_vm.count("resource")) {
+        io::odstream{tp, _path, io::leaf_resource_name{_vm["resource"].as<std::string>()}};
+        return;
+    }
+
+    if (std::strlen(_env.rodsDefResource) == 0) {
```
korydraughn, line 311: This needs to be revisited. There is likely a much better/safer way to test for empty string.

trel: can't the server determine where to create the data object in the absence of a default coming from the client?

korydraughn: That's the thing, there is no server side code to check this. `itouch` is implememted entirely on the client side. Which does mean the microservice version has to be implemented from the ground up.

We may have to create a new API plugin and move all of this into it. Then the clients can just invoke an API endpoint and everything behaves the same.

korydraughn:
This implementation will serve as the base for a new server-side API plugin.

See irods/irods#5152

trel:
closed in favor of https://github.com/irods/irods_client_icommands/pull/152

PR #151, opened 2020-09-17
New icommand: itouch (4-2-stable)

korydraughn:


src/itouch.cpp
```diff
@@ -0,0 +1,399 @@
...
+            if (vm.count("replica") || vm.count("resource")) {
+                std::cerr << "Error: -n and -R cannot be used for collections.\n";
+                return 1;
+            }
+
+            if (fs::client::is_collection_registered(conn, path)) {
+                fs::client::last_write_time(conn, path, new_mtime);
+            }
+            else {
+                // TODO Should anything happen here?
```
trel, line 144: this is... what, a collection that doesn't exist in the catalog?

src/itouch.cpp
```diff
@@ -0,0 +1,399 @@
...
+            }
+        }
+        else if (fs::client::is_data_object(object_status)) {
+            if (fs::client::is_data_object_registered(conn, path)) {
+                const auto replica_number = get_replica_number(vm, conn, path);
+                ix::replica::last_write_time<RcComm>(conn, path, replica_number, new_mtime);
+            }
+            else {
+                // TODO Should anything happen here?
```
trel, line 153: and this is a data object that doesn't exist in the catalog?   part of a special collection archive file?

src/itouch.cpp
```diff
@@ -0,0 +1,399 @@
...
+            }
+        }
+        else {
+            // TODO Should anything happen here?
```
trel, line 157: this is... what?

src/itouch.cpp
```diff
@@ -0,0 +1,399 @@
...
+
+auto print_usage_info() noexcept -> void
+{
+    std::cout << R"_(itouch - Change logical path timestamp
+
+Usage: itouch [OPTION]... LOGICAL_PATH
+
+Update the modification time of a logical path to the current time.
+
+A LOGICAL_PATH argument that does not exist is created empty, unless -c
```
trel, line 184: ...'not exist will be created as an empty data object, unless'...

src/itouch.cpp
```diff
@@ -0,0 +1,399 @@
...
+is supplied.
+
+If a replica number or leaf resource is not specified, the latest good
+replica will be updated.
+
+Mandatory arguments to long options are mandatory for short options too.
+
+Options:
+  -c, --no-create  Do not create any data objects.
```
trel, line 193: 'Do not create a data object.'

src/itouch.cpp
```diff
@@ -0,0 +1,399 @@
...
+  -n, --replica    The replica number of the replica to update.  This
+                   option applies to replicas only.  Cannot be used
```
trel, line 195: 'replicas' -> 'data objects'

src/itouch.cpp
```diff
@@ -0,0 +1,399 @@
...
+                   with -R.
+  -R, --resource   The leaf resource that contians the replica to update.
+                   This option applies to replicas only.  Cannot be
```
trel, line 198: 'replicas' -> 'data objects'

trel (commented):
do we need to consider an admin flag?   -M?    can be a new issue for later perhaps.

src/itouch.cpp
korydraughn: yes

src/itouch.cpp
korydraughn: yes, it doesn't exist in the catalog.  likely part of a special collection

src/itouch.cpp
korydraughn: I'm not sure. I'd hope this case is unreachable.

src/itouch.cpp
korydraughn: maybe we print a message saying the mtime cannot be updated.

src/itouch.cpp
korydraughn: maybe we print a message saying the mtime cannot be updated.

korydraughn:
What is the use-case for an admin flag?

trel:
So a rodsadmin can touch a data object / collection that they do not own or have write access to.

Could be used to trigger other policy that cares about mtimes and takes action (data movement, caching, etc.).

Definitely can be a new enhancement issue.

src/itouch.cpp
trel: yes, perhaps `LOG_WARNING`, as this shouldn't be reached in normal operation?  or just `LOG_NOTICE`?

src/itouch.cpp
trel: seems `LOG_NOTICE` is good here... it's not a warning, unless we're asserting they shouldn't have gotten into this codepath.

src/itouch.cpp
trel: definitely a LOG_ERROR - should not be here.

trel:
Please make a new issue to morph this into an API plugin.   Can get this work merged.

korydraughn:
This implementation will serve as the base for a new server-side API plugin.

See irods/irods#5152

korydraughn:
After porting some of this code into the server, I'm starting to believe that continuing on that path is incorrect. `touch` makes sense from the client side, but not on the server side. There are things that are needed by `touch` that are available on the client side, but not on the server side (e.g. info from `irods_environment.json`). I feel that this will turn into something that implements half of the options in the server (for no real benefit) and require clients to implement the remaining options. We have demonstrated that `touch` can be implemented to some extent. I think other tools should use this implementation as a reference for adding touch-like behavior.

Thoughts?

korydraughn:
Maybe we can make this work. See https://docs.irods.org/4.2.8/system_overview/configuration/#default-resource-configuration (thanks @trel).

PR #155, opened 2020-10-07
[irods/irods#5111] Implement long output for non-data objects in imeta ls (master)

SwooshyCueb:
This is meant to address irods/irods#5111.

Test will be in a separate pull request in the irods repository.


trel:
looks good - please add the #

SwooshyCueb:
Added the #

PR #156, opened 2020-10-08
[irods/irods#5111] Implement long output for non-data objects in imeta ls (4-2-stable)

SwooshyCueb:


trel:
looks good - please #

SwooshyCueb:
Added the #

PR #163, opened 2020-10-21
[irods/irods#5081] imeta interactive: do not exit on error status from do_command (master)

SwooshyCueb:
Addresses irods/irods#5081

### DIFF
--- a/src/imeta.cpp
+++ b/src/imeta.cpp
@@ -1189,8 +1189,10 @@ getInput( char *cmdToken[], int maxTokens ) {
     char *cpTokenStart;
     char *stat;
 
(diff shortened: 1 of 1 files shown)


trel:
seems pretty straightforward.

an error from imeta will now be printed, but not make imeta exit, is that correct?

SwooshyCueb:
That's correct. The only return value from do_command that will cause an exit is `-1`, which is what's returned on the `quit` command.

trel:
excellent.   please # em.

SwooshyCueb:
All tests passed except the one that doesn't. #s added

PR #164, opened 2020-10-21
[irods/irods#5081] imeta interactive: do not exit on error status from do_command (4-2-stable)

SwooshyCueb:
#163 cherry-picked into stable.

Addresses irods/irods#5081

Currently being pushed through jenkins, will update when done


SwooshyCueb:
All tests passed except the one that doesn't

PR #165, opened 2020-10-23
[irods/irods#5214] iuserinfo: fix use after free in main (master)

SwooshyCueb:
The data backing the pointer returned by `std::string::c_str()` only sticks around as long as the `std::string` object. In this case, the `std::string` object is only around long enough to return the pointer.

This also fixes a warning from clang 10.

Addresses irods/irods#5214

### DIFF
--- a/src/iuserinfo.cpp
+++ b/src/iuserinfo.cpp
@@ -159,9 +159,9 @@ main( int argc, char **argv ) {
         status = showUser(argv[myRodsArgs.optind]);
     }
     else {
-        const auto user_name{(boost::format("%s#%s") %
-                              myEnv.rodsUserName % myEnv.rodsZone).str().c_str()};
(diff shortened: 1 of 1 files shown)


SwooshyCueb:
Waiting for comments before I push through jenkins

korydraughn (approved):

trel:
we'll need an issue for this in the main repo - and then please point at it - # whenever ready

SwooshyCueb:
# added

trel:
iords in the commit - please tweak

SwooshyCueb:
I ought to put together a post-commit hook that does a spellcheck for me. I think that's the second time I've bungled a commit message.
Fixed.

trel:
not a problem - why we have a process :)

PR #166, opened 2020-10-26
[irods/irods#5214] iuserinfo: fix use after free in main (4-2-stable)

SwooshyCueb:
Cherry-pick of #165

The data backing the pointer returned by `std::string::c_str()` only sticks around as long as the `std::string` object. In this case, the `std::string` object is only around long enough to return the pointer.

This also fixes a warning from clang 10.

Addresses irods/irods#5214


trel:
yep, please # it

SwooshyCueb:
#'d

PR #168, opened 2020-11-04
[irods/irods#5220] cmake: use GNUInstallDirs (master)

SwooshyCueb:
Addresses irods/irods#5220 for this repository.

Will wait for feedback before pushing through Jenkins.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -14,14 +14,11 @@ set(CMAKE_EXPORT_COMPILE_COMMANDS TRUE)
 
 include(${IRODS_TARGETS_PATH})
 
+include(GNUInstallDirs)
+
 if (NOT CMAKE_CONFIGURATION_TYPES AND NOT CMAKE_BUILD_TYPE)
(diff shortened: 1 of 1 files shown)


trel:
any way to reuse/share the cmake/InstallDirs.cmake from main irods/irods?    no worries, it's not big, but if it's identical...

is that a bad idea?      

please run through CI

SwooshyCueb:
Since the icommands are designed to be built without the irods sources present, the only practical way I know of to do this is to install `InstallDirs.cmake` into the install prefix somewhere and maybe point to it in one of the `lib/irods/cmake` files. Considering how small it is, it's probably more trouble than it's worth, at least for the time being.

trel:
Agreed.   Trying to get them decoupled AND not duplicate code are sometimes competing interests...

trel:
Don't we already use some CMake magic from the irods-dev package?    We can add to the magic?

trel:
yes, looking very nice.

PR #175, opened 2020-12-10
[irods/irods#5294] istream honors client-side configured default resource on writes (master)

korydraughn:


trel:
seems good - need a test in irods/irods i suppose - # when ready

korydraughn:
added #

PR #185, opened 2021-01-07
[irods/irods#4761] Added support for case-insensitive search to ilocate (4-2-stable)

korydraughn:


bin/ilocate
```diff
@@ -2,8 +2,6 @@
 # ilocate - simple script to emulate functionality of GNU's [m]locate command for iRODS
 # Authored by Scott Sinno, NASA NCCS, scott.s.sinno@nasa.gov
 #
-#Todo : 
-#	Add functionality for case-insensitive search, preferably by way of 
 # adding support for it in iquest.
```
trel, line 5: missed a line...

bin/ilocate
```diff
@@ -20,21 +18,29 @@ then
...
-		then
-			echo "$sql_output" | grep -v "CAT_NO_ROWS_FOUND"
-			return 0
-		else
-			return 1
-		fi
+        if [ $case_insensitive -eq 0 ]; then
+		sql_output=`iquest --no-page "%s/%s" "select COLL_NAME, DATA_NAME WHERE $1 $operand '$2'" 2> /dev/null`
+        else
+		sql_output=`iquest uppercase --no-page "%s/%s" "select COLL_NAME, DATA_NAME WHERE $1 $operand '${2^^}'" 2> /dev/null`
```
trel, line 34: this requires bash 4+?   please confirm that's shipped with our supported OS/versions.

korydraughn: replaced the solution with a bash 3 impl. the `awk` impl should be safe.

thoughts?

korydraughn:
Just added `ilocate` to `ihelp`. I think that is all that is needed to make `ilocate` appear in the docs?

bin/ilocate
trel: seems good.

we have a test?

trel:
i believe that's true - the docs are generated from `ihelp -a`

korydraughn:
The formatting for this file is off. I'll expand everything to spaces and add another commit for the same issue.

bin/ilocate
korydraughn: i didn't see any tests for it. i'll look again.

korydraughn: Added test.

korydraughn:
Added #

PR #187, opened 2021-01-27
[irods/irods#5250] cmake: set version number in project call (master)

SwooshyCueb:
in service of irods/irods#5250


trel (commented):
cmake minimum version should be 3.7.0 i think...

SwooshyCueb:
> cmake minimum version should be 3.7.0 i think...

Bumped it up

PR #189, opened 2021-01-31
[irods/irods_testing_jenkins#116] "gzip -k" alternative for CentOS 7 (master)

korydraughn:
For more info, see https://github.com/irods/irods_testing_jenkins/issues/116


korydraughn:
@SwooshyCueb Any concerns around these changes?

SwooshyCueb (approved):
I typically try to avoid using anything that requires a shell in cmake, but if you've tested this and it works fine, I've got no issue merging this.

korydraughn:
added #

trel:
Yes, this doesn't add a new shell command... just manipulates one we've already got.

Removing shell is the future... but not today.

SwooshyCueb:
> Yes, this doesn't add a new shell command... just manipulates one we've already got.

What I mean to say is that I try to avoid assuming this kind of stuff runs in a shell to begin with vs being executed directly (`execvp` style) by make or ninja or whathaveyou.
But again, I'm okay with this being merged if it's been tested.

PR #194, opened 2021-03-09
[#5477] istream honors the size in the catalog on reads (master)

korydraughn:


trel:
commit message should point to [irods/irods5477]

korydraughn:
passed CI. added #

PR #196, opened 2021-03-16
[irods/irods#5082] Introduce userspace tarball packaging functionality (master)

SwooshyCueb:
Introduce packaging functionality for a userspace tarball.
Addresses irods/irods#5082

The current implementation is separate from our CPack packaging.

Requires Python 3. I'm targeting 3.6/3.5 for master/4-2-stable, which should ensure compatibility with all distros we currently support.
Will use `lief` if it is available, and will scream if it's not.

Here's the current process:
1. CMake adds the `userspace-tarball` target to the makefile. This target invokes the `userspace-packager.py` script and passes in a ton of information.
2. The script invokes the installation makefile targets for the icommands, with the prefix set to a temporary workspace.
3. The icommands are stripped of unneeded symbols, and, if `lief` is available, any imported libraries that are no longer needed after stripping are removed from the dynamic entry table.
4. The iRODS runtime libraries exposed by the CMake configuration file in the `irods-dev` package are copied to the temporary workspace. They are stripped and cleaned just like the icommands.
5. The script attempts to collect the needed auth and network plugins. These libraries are not exposed by the CMake configuration file in the `irods-dev` package, so the script takes a really good educated guess as to their location and finds them in the filesystem. They are stripped and cleaned just like the icommands and iRODS runtime libraries.
6. The script will attempt to identify needed libraries that we provide (irods-externals). If `lief` is available, It will collect, strip, and clean direct dependencies of everything prepared so far, repeating until no more irods-externals are required. If `lief` is not available, the script will collect all irods-externals in one go.
7. The script will attempt to collect needed libraries that are provided by the distribution but not installed by default. This is where the `known_libs` files come in. Because we cannot detect at build-time whether or not a given library is installed by default, these files contain the sonames of any libraries that I have ever seen in the dependency tree for the icommands package and a marker indicating whether or not it is okay to include them in the userspace tarball. Distribution-provided libraries are usually built with `-Wl,--as-needed`, meaning it is not as important to strip and clean them as it is the binaries we provide. At the moment, I'm planning on doing it anyway.
8. If usable `lief`, `chrpath`, or `cmake` is available, the script will set the runpath of all the prepared binaries to use a relative path to point at the lib directory inside the package. If `lief` is available, it will also set the `DF_ORIGIN` flag. This ensures that the user won't need to set `LD_LIBRARY_PATH` when running any of the icommands. the `DF_ORIGIN` flag is necessary to tell the runtime linker to respect `$ORIGIN` in the runpath, even when the libraries are loaded via `libdl` (like our plugins, for example).
9. The tarball is created.

Here's what's left to be done:
- [x] Jenkins testing
- [x] Merge related PR in irods repo
- [x] Make sure tarballs build in docker containers for our supported distros
    - [x] Ubuntu Bionic
    - [x] Centos 7
- [x] Make sure tarballs work in vanilla docker containers for our supported distros
    - [x] Ubuntu Bionic
    - [x] Centos 7
- [x] Write a little build documentation or example script
- [x] Cleanup
- [x] Finish #212
- [x] Hack together tarballs for 4.2.8

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1,16 +1,26 @@
 cmake_minimum_required(VERSION 3.7.0 FATAL_ERROR) #CPACK_DEBIAN_<COMPONENT>_PACKAGE_NAME
 
 find_package(IRODS 4.3.0 EXACT REQUIRED CONFIG)
+set(IRODS_PACKAGE_PREFIX "${PACKAGE_PREFIX_DIR}")
 
 include(RequireOutOfSourceBuild)
 
-set(CMAKE_C_COMPILER ${IRODS_EXTERNALS_FULLPATH_CLANG}/bin/clang)
 set(CMAKE_CXX_COMPILER ${IRODS_EXTERNALS_FULLPATH_CLANG}/bin/clang++)
 set(CMAKE_CXX_STANDARD 17)
+# export-dynamic so stacktrace entries from executables have function names
+set(CMAKE_EXE_LINKER_FLAGS_INIT "-rdynamic -Wl,--export-dynamic -Wl,--enable-new-dtags -Wl,--as-needed -Wl,-z,defs")
+set(CMAKE_EXE_LINKER_FLAGS_RELEASE_INIT "-Wl,--gc-sections -Wl,-z,combreloc")
 
 project(icommands
   VERSION "${IRODS_VERSION}"
-  LANGUAGES C CXX)
+  LANGUAGES CXX)
+
+set(CMAKE_SKIP_BUILD_RPATH OFF)
+set(CMAKE_SKIP_INSTALL_RPATH OFF)
+set(CMAKE_SKIP_RPATH OFF)
+set(CMAKE_INSTALL_RPATH_USE_LINK_PATH ON)
+set(CMAKE_BUILD_WITH_INSTALL_RPATH OFF)
+set(CMAKE_BUILD_RPATH_USE_ORIGIN ON)
 
 include(UseLibCXX)
 
@@ -97,6 +107,8 @@ endif()
 
 set(CMAKE_INSTALL_RPATH_USE_LINK_PATH TRUE)
 add_compile_options(-Wall -Wextra -Werror -Wno-unused-parameter)
+
+add_custom_target(icommands)
 
 set(
   IRODS_CLIENT_ICOMMANDS_EXECUTABLES
@@ -185,11 +197,18 @@ foreach(EXECUTABLE ${IRODS_CLIENT_ICOMMANDS_EXECUTABLES})
     )
   target_compile_definitions(${EXECUTABLE} PRIVATE ${IRODS_COMPILE_DEFINITIONS} BOOST_SYSTEM_NO_DEPRECATED)
   target_compile_options(${EXECUTABLE} PRIVATE -Wno-write-strings)
+  if (CMAKE_VERSION VERSION_LESS "3.13.0")
+    set_property(TARGET ${EXECUTABLE} APPEND PROPERTY LINK_FLAGS "-Wl,-z,origin")
+  else()
+    target_link_options(${EXECUTABLE} PRIVATE "LINKER:-z,origin")
+  endif()
+  add_dependencies(icommands ${EXECUTABLE})
   install(
     TARGETS
     ${EXECUTABLE}
     RUNTIME
     DESTINATION ${CMAKE_INSTALL_BINDIR}
+    COMPONENT icommands
     )
 
   if (IRODS_LINUX_DISTRIBUTION_NAME STREQUAL "centos")
@@ -222,6 +241,7 @@ foreach(EXECUTABLE ${IRODS_CLIENT_ICOMMANDS_EXECUTABLES})
       FILES
       ${EXECUTABLE_MANPAGE}.gz
       DESTINATION ${CMAKE_INSTALL_MANDIR}/man1
+      COMPONENT manpages
       )
   endif()
 endforeach()
@@ -238,6 +258,7 @@ foreach(IRODS_CLIENT_ICOMMANDS_SCRIPT ${IRODS_CLIENT_ICOMMANDS_SCRIPTS})
     ${SCRIPT_SRC_PATH}
     DESTINATION ${CMAKE_INSTALL_BINDIR}
     PERMISSIONS OWNER_READ OWNER_WRITE OWNER_EXECUTE GROUP_READ GROUP_EXECUTE WORLD_READ WORLD_EXECUTE
+    COMPONENT icommands
     )
 
   if (IRODS_LINUX_DISTRIBUTION_NAME STREQUAL "centos")
@@ -266,6 +287,7 @@ foreach(IRODS_CLIENT_ICOMMANDS_SCRIPT ${IRODS_CLIENT_ICOMMANDS_SCRIPTS})
       FILES
       ${SCRIPT_MANPAGE}.gz
       DESTINATION ${CMAKE_INSTALL_MANDIR}/man1
+      COMPONENT manpages
       )
   endif()
 endforeach()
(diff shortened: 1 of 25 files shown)


packaging/userspace/userspace-packager.py
```diff
@@ -0,0 +1,331 @@
...
+	if not is_bsdtar and not is_gnutar:
+		raise OSError(errno.ENOPKG, "tar tool incompatible (only GNU tar and bsdtar are supported)")
+
+	tar_args = [tar_path, '-c', '-a', '-f', options.output_path]
+
+	if is_bsdtar:
+		tar_args.append('--no-fflags')
+		# bsdtar will derive format from filename
+		tar_args.extend(['--numeric-owner', '--uid', '0', '--gid', '0'])
+	elif bsdtar:
```
trel, line 53: is_gnutar?

packaging/userspace/userspace-packager.py
```diff
@@ -0,0 +1,331 @@
...
+	os.mkdir(options.work_dir)
+
+	pkg_dir = path.join(options.work_dir, 'pkg')
+	os.mkdir(pkg_dir)
+
+	icommands_paths = prepare_icommands(options, pkg_dir)
+	iruntime_paths = prepare_irods_runtime(options, pkg_dir)
+
+	# TODO:
+	# 1. Prepare irods extenrs
```
trel, line 315: took me longer than it should have to realize this was supposed to be externs

packaging/userspace/userspace-packager.py
```diff
@@ -301,18 +309,33 @@ def _main():
...
+		l.warning('lief cannot be imported. Will not be able to remove DT_NEEDED entries for ' +
+		          'unneeded libraries or set DF_ORIGIN flags. This can result in additional ' +
+		          'libraries being included in the package, and may also result in ' +
+		          'LD_LIBRARY_PATH needing to be set to run programs from the package.')
+		if options.chrpath_path is None:
+			options.chrpath_path = shutil.which('chrpath', path=env['PATH'])
+			if options.chrpath_path is None:
+				l.error('chrpath tool cannot be found. Without either lief or chrpath, will ' +
+				        'not be able to set rpath/runpath in package binaries. This will ' +
+				        'result in bad runpath/rpath values and LD_LIBRARY_PATH needing to be ' +
```
trel, line 322: you switched the order of rpath and runpath in these two lines.   pick an order, run with it.

packaging/userspace/externs_libs.common
```diff
@@ -0,0 +1,34 @@
...
+# (irods-externals) in library dependency chains.
+#
+# Each line lists the name of a library with no file extension or soversion.
+# Blank lines are ignored.
+# Lines starting with # are ignored.
+#
+# RATIONALE: In order to support building the userspace package when irods or the icommands are
+# built against irods-externals in nonstandard locations, a list of library names is provided
+# to assist in identifying irods-externals libraries during packaging. The data in this file will
+# only be used if the pacakger detects a dependency on a library outside of /opt/irods-externals
```
trel, line 11: pacakger -> packager

packaging/userspace/script_common.py
```diff
@@ -8,35 +8,138 @@
...
 
-def augment_module_search_paths():
-	if not path.isabs(irods_script_dir):
-		raise OSError(errno.EINVAL, "irods_script_dir is not an absoltue path", irods_script_dir)
-	if not path.exists(irods_script_dir):
-		raise OSError(errno.ENOENT, "irods_script_dir not found", irods_script_dir)
-	if not path.isdir(irods_script_dir):
-		raise OSError(errno.ENOTDIR, "irods_script_dir is not a directory", irods_script_dir)
+	if not path.isabs(irods_scripts_dir):
+		raise OSError(errno.EINVAL, "irods_scripts_dir is not an absoltue path", irods_scripts_dir)
```
trel, line 46: absoltue -> absolute

packaging/userspace/script_common.py
```diff
@@ -8,35 +8,138 @@
...
+	l = logging.getLogger(__name__)
+
+	if lief_ver is None:
+		if whine:
+			l.warning('lief version could not be detected. Caveat emptor.')
+	elif lief_ver < lief_min_req_ver:
+		if whine:
+			l.warning('lief version ' + str(lief_ver) + ' detected. Version ' +
+			          str(lief_min_req_ver) + ' or greater is required to enable lief' +
+			          ' functioniality. Version ' + str(lief_min_rec_ver) +
```
trel, line 124: functioniality -> functionality

CMakeLists.txt
```diff
@@ -266,21 +280,103 @@ foreach(IRODS_CLIENT_ICOMMANDS_SCRIPT ${IRODS_CLIENT_ICOMMANDS_SCRIPTS})
...
+endforeach()
+
+file(
+  GLOB ICOMMANDS_USERSPACE_TARBALL_INSTRUMENTATION
+  LIST_DIRECTORIES false
+  "${CMAKE_CURRENT_SOURCE_DIR}/packaging/userspace/*"
+  )
+
+if(DEFINED IRODS_EXTERNALS_FULLPATH_ARCHIVE AND IS_DIRECTORY "${IRODS_EXTERNALS_FULLPATH_ARCHIVE}/bin")
+  # Make septenvigintuple sure IRODS_EXTERNALS_FULLPATH_ARCHIVE doesn't point to the system
```
trel, line 315: that's a lot

packaging/userspace/strip.py
```diff
@@ -0,0 +1,246 @@
...
+
+	os.chmod(output_path, stat.S_IRUSR | stat.S_IWUSR | stat.S_IXUSR |
+	                      stat.S_IRGRP |                stat.S_IXGRP |
+	                      stat.S_IROTH |                stat.S_IXOTH)
+
+	return output_path
+
+lib_size_warn_threshold = (36 * 1024 * 1024)
+if lief_ver_good and script_common.lief_ver < script_common.lief_min_rec_ver:
+	lib_size_warn_threshold = (24 * 1024 * 1024)
```
trel, line 74: are these known thresholds?   fine if so, but if these are our numbers, consider explaining and/or making them knobs/inputs...

SwooshyCueb:
Okay. I'm nearly there. I can see the light at the end of the tunnel.

Still remaining to be done:
- Iron out some kinks with lief
- implement symlink handling
- cleanup and testing

packaging/userspace/libdirectives.py
```diff
@@ -0,0 +1,154 @@
...
+
+	@property
+	def libnames(self):
+		return self._libnames
+
+	def __contains__(self, lib):
+		if lib in self.sonames or lib in self.libnames:
+			return True
+
+		# don't bother checking if a passed in soname maches a libname if the soname doens't
```
trel, line 44: doens't -> doesn't

packaging/userspace/runpath_util.py
```diff
@@ -0,0 +1,231 @@
...
+				ent_flags_1 = lib.get(lief.ELF.DYNAMIC_TAGS.FLAGS_1)
+				ent_flags_1.add(lief.ELF.DYNAMIC_FLAGS_1.ORIGIN)
+			else:
+				ent_flags_1 = lief.ELF.DynamicEntryFlags(lief.ELF.DYNAMIC_TAGS.FLAGS_1,
+					int(lief.ELF.DYNAMIC_FLAGS_1.ORIGIN))
+				lib.add(ent_flags_1)
+
+		# if we have rpath but no runpath, change rpath to runpath
+		if lib.has(lief.ELF.DYNAMIC_TAGS.RPATH) and not lib.has(lief.ELF.DYNAMIC_TAGS.RUNPATH):
+			lib.get(lief.ELF.DYNAMIC_TAGS.RPATH).tag = lief.ELF.DYNAMIC_TAGS.RUNPATH
```
trel, line 84: just a check ... is this backwards?   a quick read expected it the other way around.

packaging/userspace/script_common.py
```diff
@@ -0,0 +1,248 @@
...
+	ipathargs.add_argument('--irods-install-prefix', metavar='PATH',
+		action='store', dest='irods_install_prefix',
+		type=str, default='/usr')
+	ipathargs.add_argument('--irods-package-prefix', metavar='PATH',
+		action='store', dest='irods_package_prefix',
+		type=str, default=None)
+	ipathargs.add_argument('--irods-home-dir', metavar='PATH',
+		action='store', dest='irods_home_dir',
+		type=str, default='/var/lib/irods',
+		help='iRODS scripts directory')
```
trel, line 224: iRODS home directory

packaging/userspace/userspace-packager.py
```diff
@@ -0,0 +1,521 @@
...
+		# soname : path
+		self.prepared_libs = dict()
+
+		# mapping of soname to path for all external libraries we depend on (pre-strip)
+		self.ext_lib_paths = dict()
+
+		# virtual libs (vdso), ld itself, etc.
+		self.ldd_other_libs = set()
+
+		# libraries encountered during dependancy resolution that do not have a directive
```
trel, line 58: dependancy  -> dependency

SwooshyCueb:
Update on where I'm at with this, since I haven't pushed anything up in a bit:

Symlink handling is done. We can now deal with some limited amount of symlink chicanery when gathering libraires to package.
In the case of libc++, however, `libc++.so` is not a symlink, but a linker script file. I decided that instead of pulling in a dependency or implementing a parser to deal with this, we can avoid it by substituting in the versioned soname in that specific case. The libc++ soname version hasn't changed in its entire lifespan, so I think it's safe to assume that it will continue to not change for as long as my packager is relevant.

I've done a good amount of cleanup and reorganization, and documented a few places where I have some low-priority TODOs. `flake8` is coming in very handy here. In order to make life easier for myself, I've also added type-hinting here and there.

I've started doing testing in docker with the distros we support.

I'm still ironing out some kinks with lief. Sometimes it does some things the runtime linker doesn't like. I've decided that, even though it's *very* useful, it should be avoided when alternatives are available. In service of this, I've started (and mostly completed) work on a third way of setting RUNPATH/RPATH, using `cmake -P`. It works in much the same way as `chrpath`, but doesn't require the installation of an additional utility. Additionally, for the icommands, the `DF_ORIGIN` flag is now set as part of the build process. I've done this by passing an extra argument to the linker. I'll be submitting a companion pull request to have this done for the irods runtime libraries and plugins as well.

On the topic of `DF_ORIGIN`, I have discovered that handling of the `$ORIGIN`/`${ORIGIN}` substitution string and the `DF_ORIGIN` flag is not done according to the spec in the vast majority of cases. I'm still poking and prodding to figure out *when* exactly we need to have this flag set, as setting it requires the use of lief (with the exception of the areas where we can do it ourselves).

On the topic of RUNPATH/RPATH, it turns out that neither `chrpath` nor CMake can *add* entries to the dynamic section of an ELF, so if a binary does not already have a set RUNPATH/RPATH, `lief` will be required to set one. (We do run into this with nanodbc.) As a result, it no longer makes sense to pass multiple binaries to `chrpath` at once (when it is determined that the `chrpath` method makes the most sense), so I've started re-working `RunpathUtil` to not do this, and to fall back to lief if `chrpath` fails and lief is available.

To reiterate a couple of things I've said in an email chain, this packager will only work for what's currently in git. I will be back-porting it for stable, but for anything currently released, it won't apply. I will, however, be manually creating tarballs for 4.2.8 once I've got this PR merged. Additionally, there aren't any functional changes to the way the existing buildsystem works for icommands or the irods runtime, so any homegrown packaging efforts built on top of the existing infrastructure should be unaffected.

If you'll pardon the mixed metaphor, I'm in the home stretch, but still peeling back the layers, it seems.

packaging/userspace/runpath_util.py
SwooshyCueb: We prefer `RUNPATH` over `RPATH`, so unless I'm missing something, this is correct.

trel: i see it now - definitely read it backwards myself.

SwooshyCueb:
Just did another big push. Since I squashed all my commits before pushing, here's a list of notable changes since the last push:
- icommands are now built with the origin flag by default. This eliminates the need to use LIEF to set their runpaths under most conditions (something which previously had the potential to bungle up the files)
- icommands are now linked with `--gc-sections` and `-z combreloc` by default for release builds. This makes things easier for LIEF
- icommands are now linked with `--as-needed` by default. This tells the linker to ignore passed-in libraries from which no symbols are actually referenced.
- C is removed from the list of languages in the CMake `project` call. Everything here is C++.
- A `.flake8` configuration file is included.
- I ran `flake8` over all the python files and did a hefty amount of cleanup. Also, type annotations!
- I did a hefty amount of reorganization.
    - A lot of stuff that used to live in `script_common.py` now lives in `context.py`, `log_instrumentation.py`, and `options.py`
    - Since we're supporting multiple versions of Python, all the way back to Xenial's 3.5.2 (which is actually 3.5.1 pretending to be 3.5.2), I've added a `compat_shims.py` module that bridges gaps where needed.
        - NOTE: The module includes a shim for `typing.overload`, which is broken in 3.5.0 and 3.5.1. Since `overload` is no longer imported directly from `typing`, most linters fail to recognise the decordator and will complain about method redefinition when it is used. (Codacy *does* get very angry about this.) This can be safely ignored.
        - NOTE: For some reason, some linters think that `typing.OrderedDict` doesn't exist, even in >3.7.2.
        - NOTE: For some reason, under certain conditions, some linters crash on the first line where `collections.OrderedDict` is directly referenced. Watch out.
    - `ldd_util.py` is now `elfinfo_util.py` and has more functionality.
        - The functionality for determining the direct dependencies of an ELF have moved from `userspace-packager.py` to the elfinfo utility. It was also expanded to provide more info:
            - rpath
            - runpath
            - direct dependencies
            - presence of origin flag in `DT_FLAGS` entry
            - presence of origin flag in `DT_FLAGS_1` entry
            - list of all direct dependencies from which versioned symbols are imported, and the relevant version names
        - The elfinfo utility can now use `readelf` in addition to LIEF to determine most of the information listed above. Regexes to get most of the rest (and to get the same info from `objdump`) are included for future use.
    - The instrumentation for argument parsing has changed significantly.
    - Tool path resolution now happens when the utility classes are initialized, rather than at each run of the tool.
    - The runpath utility class is reworked to track available and attempted methods (method here meaning approach) so that, should the preferred method fail, the utility can fall back to another method.
    - For multi-part single-destination operations, instead of using the destination path for the output of the first part of the operation and then performing the rest of the operation in-place, intermediate files are created in `${CMAKE_CURRENT_BINARY_DIR}${CMAKE_FILES_DIRECTORY}/CMakeTmp`
- The runpath utility is now slightly smarter about determining when LIEF is needed and when LIEF cannot be used.
- The runpath utility can now use `CMake -P` with the `chrpath.cmake` script to set runpath/rpath if `chrpath` fails or cannot be found
- The directory structure of the tarball will now follow the `CMAKE_INSTALL_BINDIR`, `IRODS_INSTALL_LIBDIR`, and `IRODS_PLUGINS_DIRECTORY` CMake variables.
- The packager can now handle a limited amount of symlink chicanery.
- The default tarball filename now includes the icommands build configuration name.

Additionally, I've submitted irods/irods#5573, which fixes some bad assumptions regarding default paths in which irods components are stored on the system. This is required in order for the tarballs to have a mundane directory structure.

And now a question:
The irods runtime packages contain some plugins:
- `auth/libnative_client.so`
- `auth/libnative_server.so`
- `auth/libosauth_client.so`
- `auth/libosauth_server.so`
- `auth/libpam_client.so`
- `auth/libpam_server.so`
- `network/libssl_client.so`
- `network/libssl_server.so`
- `network/libtcp_client.so`
- `network/libtcp_server.so`

Can I safely exclude the `*_server.so` plugins from the tarballs?

trel:
> Can I safely exclude the *_server.so plugins from the tarballs?

yes.

trel:
so the 'test' for this...
- someone else builds this thing
- sends the resulting tarball to a different machine
- explodes the tarball
- runs `iinit` against a known remote iRODS system
- profit!

is that right?   anything missing?  do we have a client-only iCommands suite to put this through its paces?

trel:
Since we know that the iCommands themselves are as good as the SHA they come from... we don't need to actually exercise all their functionality for a tarball test suite.

What is the minimum thing we need to verify?  Just that they're linked correctly?  A clean walk of all the binaries with `ldd`?

SwooshyCueb:
So here's my current agenda:
- Exclude the `*_server.so` plugins from the tarball.
- Bridge the gap between "works on my machine" and "builds in jenkins" for irods/irods#5573 and this PR. Something I changed somewhere has made CMake angry.
- Make sure everything builds in docker containers for our supported distros
- Make sure the extracted tarballs work on vanilla docker images for our supported distros
- Write a little build documentation
- Backport everything to stable and do it all again
- Hack together tarballs for 4.2.8

I don't know if there's a relevant test suite or not (I'm betting not), but I do know if we throw the regular test suite at it, things will go very bad very quickly. (Ask me how I know!)

To make sure that everything links correctly and none of the libraries have been bungled in a way `ldd` won't pick up on (it's happened), we should probably test running each one once. I imagine that for many of them, just getting the help output would be enough.

trel:
good plan.

i'm fine with a simple script to run each command once... however, help text will be local only, no network traffic, so may not be sufficient?  will continue to brainstorm.

packaging/userspace/build_and_package.example.sh
```diff
@@ -0,0 +1,62 @@
...
+ICOMMANDS_SRC_DIR="$(readlink -s -m "${ICOMMANDS_SRC_DIR:-"${SCRIPTDIR}/../.."}")" # icommands source directory
+ICOMMANDS_BLD_DIR="$(readlink -s -m "${ICOMMANDS_BLD_DIR:-"${ICOMMANDS_SRC_DIR}/../icommands_build/userspace"}")"
+
+IRODS_SRC_DIR="$(readlink -s -m "${IRODS_SRC_DIR:-"${ICOMMANDS_SRC_DIR}/../irods_source"}")" # irods source directory
+IRODS_BLD_DIR="$(readlink -s -m "${IRODS_BLD_DIR:-"${ICOMMANDS_BLD_DIR}/irods_build"}")" # irods build directory
+
+IRODS_BUILD_PREFIX="$(readlink -s -m "${IRODS_BUILD_PREFIX:-"${IRODS_BLD_DIR}/pfx"}")" # temporary install prefix
+
+QTY_CPUS="$(nproc --all)"
+QTY_JOBS="${QTY_JOBS:-"$((QTY_CPUS * 2 / 3))"}"
```
trel, line 17: for 1 CPU... QTY_JOBS becomes 0?

why 2/3?     not a big deal, just curious.

packaging/userspace/build_and_package.example.sh
```diff
@@ -0,0 +1,62 @@
...
+
+CMAKE="${CMAKE:-cmake}"
+IRODS_BUILD_CONFIGURATION="${BUILD_CONFIGURATION:-Release}"
+
+
+## Prepare workspace
+
+if [ ! -e "${ICOMMANDS_BLD_DIR}" ]; then
```
trel, line 25: -d ?    for all three of these?

src/itouch.cpp
```diff
@@ -1,8 +1,8 @@
-#include <irods/rodsClient.h>
-#include <irods/client_connection.hpp>
-#include <irods/irods_version.h>
-#include <irods/touch.h>
-#include <irods/irods_exception.hpp>
+#include "rodsClient.h"
```
trel, line 1: @korydraughn do you know some reason for the angle brackets?

trel: oh i see the commit message... the non-default search paths...

packaging/userspace/build_and_package.example.sh
SwooshyCueb: I usually run about 14 or 16 on my system, which has 24 cores. The idea is to use most cores, but not all, so the user can still make use of their machine while the build happens.

src/itouch.cpp
SwooshyCueb: Yeah, `#include <irods/*>` will only work if all the headers are in a folder named `irods` that's in a system include search path. The include path we pass to the compiler already has the `irods` bit, so it would be skipped over while searching (assuming it does count as a system include search path).

SwooshyCueb:
Works for Centos 7 now. That said, I haven't tested every individual icommand for either distro.

Some details about packages required to use the packager:
- Both Centos 7 and Ubuntu 18.04 need the `python3-packaging` package installed via yum/apt. Both also need the `lief` pip package. Easiest way to get this is to install `python3-pip` via yum/apt then run `pip3 install lief`.
- Since CMake-based runpath setting is broken currently for Centos 7, the `chrpath` yum package should also be installed. I haven't tested this on Ubuntu 18.04, as `chrpath` is already present in my builder docker image. (I think our dependencies already pull in whichever package contains `chrpath` on Ubuntu.)
- If the packager is being run with Python 3.8 or newer, the `distro` module is required. Both Centos 7 and Ubuntu 18.04 use Python 3.6, so this shouldn't come up for a bit.

SwooshyCueb:
Squashed and ready for final review.

trel:
I like it - please # if you're happy.   This is master, so we can tweak it later as well (in a new issue) if we find something else.

SwooshyCueb:
\#'d

PR #197, opened 2021-03-25
[irods/irods#5506] Added option to ils that skips printing collection's contents (master)

korydraughn:
Depends on https://github.com/irods/irods/pull/5509


korydraughn:
added #

PR #198, opened 2021-03-25
[irods/irods#5506] Added option to ils that skips printing collection's contents (4-2-stable)

korydraughn:
Depends on https://github.com/irods/irods/pull/5510


korydraughn:
passed CI

added #

PR #199, opened 2021-04-04
Allow quoted arguments in imeta interactive mode (4-2-stable)

d-w-moore:
Fixes [irods #5518](http://github.com/irods/irods/issues/5518)


trel:
these commits should point to irods/irods_5518

PR #200, opened 2021-04-06
Allow quoted arguments in imeta interactive mode (master)

d-w-moore:


trel:
seems good - this works with the test to prove it stays in the console/prompt?

d-w-moore:
Yes, the test passed with that included.

Daniel Moore
Applications Engineer
iRODS Consortium, RENCI, UNC Chapel Hill
(919) 445-9616

________________________________

trel:
gotcha - please (squash? and) # em.

d-w-moore:
Also, "imeta>" prints exactly once despite multiple inputs if the terminal is not a tty... so that part of this and other tests passed as well.  I thought Cyverse and others might think it strange to see the "imeta>" prompt repeating endlessly on stdout with potentially noone listening....

trel:
seems reasonable - use case is pushing a file of commands into imeta, right?

d-w-moore:
Yes, `cat`ting a file of sub commands in through stdin as Tony was doing in https://github.com/irods/irods/issues/5518

trel:
indeed.   so yes, one or two commits is fine - # em up.

PR #212, opened 2021-05-11
[irods/irods#5082] Introduce userspace tarball packaging functionality (4-2-stable)

SwooshyCueb:
#196 backported to stable
Addresses irods/irods#5082

Here's what's left to be done:
- [x] Python 3.5.1 compatibility
- [x] Make sure tarballs build in docker containers for our supported distros
    - [x] Ubuntu Bionic
    - [x] Ubuntu Xenial
    - [x] Centos 7
- [x] Make sure tarballs work in vanilla docker containers for our supported distros
    - [x] Ubuntu Bionic
    - [x] Ubuntu Xenial
    - [x] Centos 7

### DIFF
--- a/README.md
+++ b/README.md
@@ -34,3 +34,34 @@ Our dependency chain will shorten as older distributions age out.
 
 The current setup supports new C++ features on those older distributions.
 
+## Userspace Packaging
+
+A `userspace-tarball` buildsystem target is provided to generate a userspace tarball package. This package
+will contain the iCommands and all required library dependencies.
+See `packaging/userspace/build_and_package.example.sh` for an example of how to build and package the
+iCommands for userspace deployment.
+
+### Build Dependencies
+
+The userspace packager needs a few extra packages to work properly:
+- Required: Python 3.5+.
+    - Recommended: Python 3.6+
+- Required: `setuptools` Python package.
+    - Available as `python3-setuptools` via yum/apt on Centos 7/Ubuntu.
+    - Available as `setuptools` on PyPI.
+- Required: `psutil` Python module.
+    - Available as `python3-psutil` via apt on Ubuntu.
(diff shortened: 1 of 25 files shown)


SwooshyCueb:
Some details about packages required to use the packager:
- Both Centos 7 and Ubuntu need the `python3-packaging` package installed via yum/apt. Both also need the `lief` pip package. Easiest way to get this is to install `python3-pip` via yum/apt then run `pip3 install lief`.
- Both Centos 7 and Ubuntu need the `psutil` package for Python 3. On Ubuntu, `python3-psutil` can be installed with apt, and on Centos 7, `python36-psutil` can be installed with yum.
- Since CMake-based runpath setting is broken currently for Centos 7, the `chrpath` yum package should also be installed. I haven't tested this on Ubuntu 18.04, as `chrpath` is already present in my builder docker image. (I think our dependencies already pull in whichever package contains `chrpath` on Ubuntu.)
- If the packager is being run with Python 3.8 or newer, the `distro` module is required. Both Centos 7 and Ubuntu 18.04 use Python 3.6, so this shouldn't come up for a bit.

README.md
```diff
@@ -34,3 +34,34 @@ Our dependency chain will shorten as older distributions age out.
 
 The current setup supports new C++ features on those older distributions.
 
+## Userspace Packaging
+
+A `userspace-tarball` buildsystem target is provided to generate a userspace tarball package. This package
+will contain the iCommands and all required library dependencies.
+See `packaging/userspace/build_and_package.example.sh` for an example of how to build and package the
+iCommands for usersapce deployment.
```
trel, line 42: usersapce -> usersapce

trel (commented):
looks pretty great - thanks!

SwooshyCueb:
\#'d

PR #222, opened 2021-08-24
[5786] itree command

epsilon-phase:
Related to irods/irods#5786

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -159,6 +159,7 @@ set(
   isysmeta
   iticket
   itouch
+  itree
   itrim
   iunreg
   iuserinfo
--- a/src/itree.cpp
+++ b/src/itree.cpp
@@ -1,0 +1,317 @@
+#include "rodsClient.h"
+#include "client_connection.hpp"
+#include <boost/program_options.hpp>
+#include "irods_client_api_table.hpp"
+#include "irods_pack_table.hpp"
+#include "irods_parse_command_line_options.hpp"
+#include "irods/filesystem.hpp"
+
+#include "fmt/core.h"
+#include "fmt/color.h"
+#include "json.hpp"
+
+#include <regex>
+
+namespace po = boost::program_options;
+namespace fs = irods::experimental::filesystem;
+namespace ix = irods::experimental;
+namespace json = nlohmann;
+auto print_dir(const fs::path&,
+               const po::variables_map&,
+               irods::experimental::client_connection&) -> void;
+
+auto correct_path(const po::variables_map&,const rodsEnv&)->fs::path;
+auto print_usage() -> void;
+auto get_json(const fs::path&, const po::variables_map&, ix::client_connection&, unsigned int depth) -> json::json;
+
+// ANSI escape numbers. For setting the colors.
+static const auto directory_color=fmt::color::blue;
+static const auto file_color = fmt::color::green;
+static rodsEnv env;
+static std::regex matcher;
+static std::optional<std::regex> exclude_matcher;
+static unsigned int collections = 0, objects = 0;
+static std::uintmax_t total_size = 0;
+
+
+int main(int argc, char** argv){
+    int status = getRodsEnv(&env);
+    irods::experimental::client_connection conn;
+    po::options_description desc{""};
+    po::positional_options_description pod;
+    pod.add("dir",1);
+    desc.add_options()
+        ("help,h", "Display help")
+        ("dir", po::value<std::string>(), "The initial collection to render (defaults to the current working collection)")
+        ("collections-only,c", po::bool_switch(), "Only list collections")
+        ("fullpath,f", po::bool_switch(), "print the full path of each item listed")
+        ("depth,L", po::value<int>()->default_value(1000), "Limit the depth of the listing.")
+        //This does not function because the ownership variable is initialized to the
+        //original uploader of the data_object
+        //        ("owner,o", po::bool_switch(), "Display the owner along with the collection")
+        ("pattern,P", po::value<std::string>()->default_value(".*"), "Filter files by a regexp")
+        ("size,s", po::bool_switch(),"Display the size of each data object")
+        ("ignore,I", po::value<std::string>(), "Ignore matching data objects.")
+        ("classify,F", po::bool_switch(), "Display a / at the end of listings of collections")
+        ("indent", po::value<unsigned int>()->default_value(2), "The number of spaces each level of nested collection adds.")
+        // Currently this does not work because the object_status is not initialized with the necessary information to
+        // display this properly.
+        //        ("acl,A", po::bool_switch(), "Print the access control information for each object")
+        ("json,J", po::bool_switch(), "Produce json instead of a human readable output")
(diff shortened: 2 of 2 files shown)


CMakeLists.txt
```diff
@@ -163,6 +163,7 @@ set(
   iunreg
   iuserinfo
   izonereport
+  itree
```
trel, line 166: alpha order, if you please

src/itree.cpp
```diff
@@ -0,0 +1,133 @@
...
+    pod.add("dir",1);
+    desc.add_options()
+        ("help","Display help")
+        ("dir",po::value<std::string>()->default_value(env.rodsCwd), "The collection to start from(defaults to the current working collection)")
+        ("directories_only,d", po::bool_switch(),"Only list directories")
+        ("fullpath,f",po::bool_switch(),"print the full path of each item listed")
+        ("depth,L", po::value<int>()->default_value(1000), "Limit the depth of the listing.")
+        ("color,c", po::bool_switch(), "Print in color(requires ansi terminal)");
+    if ( status < 0) {
+        std::cerr << "Irods connection error" << std::endl;
```
trel, line 39: I'm not sure this is a connection issue at this point.

trel: I think @korydraughn has had opinions about `std::endl` usage before.    too much flushing?

src/itree.cpp
```diff
@@ -0,0 +1,133 @@
...
+auto directory_depth(const fs::path& origin, const fs::path& descendant) -> int{
+    int depth = 0;
+    auto d = descendant;
+    while( origin != d ) {
+        d = d.parent_path();
+        depth++;
+    }
+    return depth;
+}
+auto print_entry(const fs::collection_entry& e, const po::variables_map& vm,int depth){
```
trel, line 67: space... `vm,int` -> `vm, int`

src/itree.cpp
```diff
@@ -0,0 +1,133 @@
...
+int main(int argc, char** argv){
+    rodsEnv env;
+    int status = getRodsEnv(&env);
+    irods::experimental::client_connection conn;
+    po::options_description desc{""};
+    po::positional_options_description pod;
+    pod.add("dir",1);
+    desc.add_options()
+        ("help","Display help")
+        ("dir",po::value<std::string>()->default_value(env.rodsCwd), "The collection to start from(defaults to the current working collection)")
```
trel, line 33: consider 'The initial collection'?  and a space before the parens

src/itree.cpp
```diff
@@ -0,0 +1,133 @@
...
+        ("directories_only,d", po::bool_switch(),"Only list directories")
+        ("fullpath,f",po::bool_switch(),"print the full path of each item listed")
+        ("depth,L", po::value<int>()->default_value(1000), "Limit the depth of the listing.")
+        ("color,c", po::bool_switch(), "Print in color(requires ansi terminal)");
```
trel, line 37: space before parens

src/itree.cpp
```diff
@@ -0,0 +1,133 @@
...
+    int status = getRodsEnv(&env);
+    irods::experimental::client_connection conn;
+    po::options_description desc{""};
+    po::positional_options_description pod;
+    pod.add("dir",1);
+    desc.add_options()
+        ("help","Display help")
+        ("dir",po::value<std::string>()->default_value(env.rodsCwd), "The collection to start from(defaults to the current working collection)")
+        ("directories_only,d", po::bool_switch(),"Only list directories")
+        ("fullpath,f",po::bool_switch(),"print the full path of each item listed")
```
trel, line 35: capital P

src/itree.cpp
```diff
@@ -0,0 +1,133 @@
...
+    rodsEnv env;
+    int status = getRodsEnv(&env);
+    irods::experimental::client_connection conn;
+    po::options_description desc{""};
+    po::positional_options_description pod;
+    pod.add("dir",1);
+    desc.add_options()
+        ("help","Display help")
+        ("dir",po::value<std::string>()->default_value(env.rodsCwd), "The collection to start from(defaults to the current working collection)")
+        ("directories_only,d", po::bool_switch(),"Only list directories")
```
trel, line 34: hrm, feel strongly both ways about this one... `d` b/c consistency with unix tools... but we have Collections...

`-c` already would collide with `color`... `-C` too subtle?  thoughts anyone?

trel (commented):
very strong start.

src/itree.cpp
korydraughn: That is correct. You don't want to use it in code that is called frequently. It will degrade performance if used in hot paths.

src/itree.cpp
korydraughn: Maybe `--color` loses its short option. `--collections-only` steals `-c`. Another solution could be that `--collections-only` doesn't get a short option. However, we define the options and there is no reason we can't use some other letter.

Also, to be more like Unix/Linux, use hyphens for long option names (to separate words).

epsilon-phase: `-C` seems like a good choice for this, it also happens to be the switch that `ls` uses to control coloring, so it's more consistent anyway.

I could also make it accept `-c` in addition to `-d`

trel:
looking through https://wtsi-npg.github.io/baton/#the-baton-cookbook ...

appears the biggest difference in baton json standard usage is... that collections are always represented as full logical paths... but data_objects are not...

@epsilon-phase can `itree` already do that?

epsilon-phase:
Hmm, would you prefer a switch for that? Or should it be the default behavior?

trel:
I think default should be the `tree` way... just the local name... but it's probably more helpful in the JSON to have the fullpaths for collections, like baton.   More actionable for the next script that interacts with this.

trel:
So... an option, yes?

`-b Baton compatibility` ?     gah.

epsilon-phase:
Here's a sample showing how the collections now have the total size accompanying themof the output as of now
image

I'll get on that right away :)

trel:
or maybe the existing `-f,--fullpath` option can *only* apply to the collections... not the data_objects.

that should be... expressive enough for other tools to be useful, *and* provide compatibility with baton.

epsilon-phase:
Do you think that should be consistent between both the json and the normal output?

trel:
let's try it, see how it feels - yes.     `--fullpath` only applies to the collections

trel:
and yes, if someone asks for fullpath... they should see that both json and regular.

but... i see how that could be weird/disjoint in regular mode...

epsilon-phase:
It's also not consistent with how `tree` does it, but that doesn't bother me too much

trel:
Ah, I see what you mean.   Definitely feels more expected to just see all fullpaths or all local/relative paths (which is how I think you already have it behaving).

trel:
please squash all these to a single commit, point at [5786]... we'll add # after it looks perfect.

trel:
please remove all the intermediate commit messages - consider writing a single awesome commit message to explain what's in here.

and please point to irods/irods_5786 rather than 4786

trel:
two last tweaks in the commit message...

"visualizatino of a collectino"
to
"visualization of a collection"


and we'll merge it!

PR #223, opened 2021-08-31
[irods/irods#4083] buildsystem support for building with GCC

SwooshyCueb:
In service of irods/irods#4083


trel:
looks good - # when happy.

SwooshyCueb:
\#'d

PR #228, opened 2021-09-20
[irods/irods#5845] include api plugins in userspace tarball (master)

SwooshyCueb:
Fixes irods/irods#5845 for master.
I haven't actually tested this against a running server yet.

My plan was to include GCC/libstdc++ compatibility changes in this PR (there might have to be special handling for libstdc++ like there is for libc++), but it turns out the problem with the GCC build that I was running into isn't actually related to packaging at all.


trel:
a delightfully small edit!    please update once it's tested a bit more.

SwooshyCueb:
Tested against a running server, everything checks out. \#'d and cherry-picked into a stable PR (#229)

PR #231, opened 2021-09-27
[irods/irods#5875] Update userspace packager inclusion directives

SwooshyCueb:
Solves irods/irods#5875 by including libcurl-gnutls and its dependency tree in the inclusion directives for bionic.
An exclusion directive for libcurl was added to centos 7 as it is installed by default.

Additionally, the directive file for xenial was removed as the packager for master does not support xenial.


trel:
this separation of concerns is pleasing to me.

SwooshyCueb:
\#'d

SwooshyCueb:
okay *now* it's \#'d

trel:
two commits - only one has the #

trel:
so squash or add another, either is fine

SwooshyCueb:
Squashed

PR #237, opened 2021-11-01
[irods/irods#4231] Remove options for iadmin (4-2-stable)

alanking:
Remove the zone option for iadmin moduser because changing a user's zone
is no longer allowed. Removed some of the username sanitizing which was
being performed by iadmin before calling the rcGeneralAdmin API. The
server should be responsible for checking these inputs.

Companion PR for irods/irods#5971


trel:
looks good to me - please # when ready

PR #238, opened 2021-11-03
[irods/irods#4231] Remove options for iadmin (master)

alanking:
Remove the zone option for iadmin moduser because changing a user's zone
is no longer allowed. Removed some of the username sanitizing which was
being performed by iadmin before calling the rcGeneralAdmin API. The
server should be responsible for checking these inputs.


trel:
seems good to me - please #

alanking:
#'d

PR #245, opened 2021-12-01
[#6022] ilsresc reports error on unknown zone name (4-2-stable)

korydraughn:
Depends on https://github.com/irods/irods/pull/6027


korydraughn:
Passed CI 

Added #

PR #250, opened 2021-12-18
Updated for Ubuntu 20 and new externals (main)

korydraughn:
The changes in this PR allow the server to be compiled against the new externals packages. The use of `static_cast<void>(var)` are temporary fixes.

Demonstrated on Ubuntu 18.04.

Depends on https://github.com/irods/irods/pull/6050


korydraughn:
added #'s

PR #257, opened 2022-01-11
[irods/irods4745] update iquest to return non-zero when CAT_NO_ROWS_FOUND

trel:


alanking (approved):
Seems good to me... @korydraughn, thoughts?

src/iquest.cpp
```diff
@@ -279,8 +279,8 @@ execAndShowSpecificQuery( rcComm_t *conn, char *sql,
     }
     status = rcSpecificQuery( conn, &specificQueryInp, &genQueryOut );
     if ( status == CAT_NO_ROWS_FOUND ) {
-        printf( "No rows found\n" );
-        return 0;
+        printf( "CAT_NO_ROWS_FOUND: Nothing was found matching your query\n" );
```
korydraughn, line 282: is this message printed twice?

i noticed it is also printed to the terminal on line 404.

alanking (changes requested):
See comments...

src/iquest.cpp
trel: oh, i see - this inner one (execAndShowSpecificQuery ) will already have printed by the time 404 gets hit... so i can just remove the 404 printf() and keep the return values correct.

korydraughn: seems reasonable

trel: done

korydraughn: is there any particular reason to prefer printing the message on line 282 rather than on line 404?

trel: i don't think so - just that it happens 'closer' to the query.

shouldn't matter where - can move if you like it 'outside'.

korydraughn: no. i was just wondering.

src/iquest.cpp
```diff
@@ -433,7 +437,7 @@ main( int argc, char **argv ) {
     if ( status < 0 ) {
         if ( status == CAT_NO_ROWS_FOUND ) {
             printf( "CAT_NO_ROWS_FOUND: Nothing was found matching your query\n" );
```
alanking, line 438: Is this also a duplicate printf?

trel:
squashin'

alanking (approved):

PR #274, opened 2022-02-21
[#6187] Marked "imeta adda" deprecated (main)

korydraughn:


src/imeta.cpp
```diff
@@ -2190,8 +2193,8 @@ void usageMain() {
         " -h This help",
         " ",
         "Commands are:",
-        " add -d|C|R|u Name AttName AttValue [AttUnits] (Add new AVU triple)",
-        " adda -d|C|R|u Name AttName AttValue [AttUnits] (Add as administrator)",
+        " add  -d|C|R|u Name AttName AttValue [AttUnits] (Add new AVU triple)",
+        " adda -d|C|R|u Name AttName AttValue [AttUnits] [Deprecated] (Add as administrator)",
```
trel, line 2197: or... we just remove it from the docs?   and line 2284 does the work to explain the -M option.

src/imeta.cpp
```diff
@@ -2292,7 +2295,7 @@ usage( const char *subOpt ) {
         }
         if ( strcmp( subOpt, "adda" ) == 0 ) {
             char *msgs[] = {
-                " adda -d|C|R|u Name AttName AttValue [AttUnits]  (Add as administrator)",
+                " adda -d|C|R|u Name AttName AttValue [AttUnits] [Deprecated] (Add as administrator)",
```
trel, line 2298: same, we just... remove this line?

src/imeta.cpp
korydraughn: sounds good

src/imeta.cpp
korydraughn: sounds good

src/imeta.cpp
korydraughn: so 4.3 will support it, we just won't document it is what you're saying?

trel: right, it will report that it is deprecated, suggest a better way, but will still execute.

korydraughn: got it

korydraughn: done

src/imeta.cpp
korydraughn: done

src/imeta.cpp
```diff
@@ -2277,27 +2276,6 @@ usage( const char *subOpt ) {
...
-        }
-        if ( strcmp( subOpt, "adda" ) == 0 ) {
-            char *msgs[] = {
-                " adda -d|C|R|u Name AttName AttValue [AttUnits]  (Add as administrator)",
-                "                                     (same as 'add' but bypasses ACLs)",
-                " ",
-                "Administrators (rodsadmin users) may use this command to add AVUs",
-                "to any dataobj, collection, resource, or user.  The syntax is the same",
-                "as 'imeta add'.",
                 ""
```
alanking, line 2279: Since I'm deprecating something as well, this has been on the mind, and I want to make sure I get it right... :)

Are we supposed to remove it from the help text, or leave it in the documentation and leave a message about its being deprecated? It seems we have gone back and forth on this over the years (see `irm -U` for example of not being removed from the help text).

src/imeta.cpp
```diff
@@ -2190,9 +2191,7 @@ void usageMain() {
         " -h This help",
         " ",
         "Commands are:",
-        " add -d|C|R|u Name AttName AttValue [AttUnits] (Add new AVU triple)",
-        " adda -d|C|R|u Name AttName AttValue [AttUnits] (Add as administrator)",
-        "                                     (same as 'add' but bypasses ACLs)",
+        " add  -d|C|R|u Name AttName AttValue [AttUnits] (Add new AVU triple)",
```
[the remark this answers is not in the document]
korydraughn: sure no problem

src/imeta.cpp
korydraughn: good question.

we haven't established a rule around this just yet. this is probably the right time to do so.

for example, below are the potential rules for deprecating features of an icommand (based on this PR):
- remove references to deprecated options from help text
- server should add deprecation message to rError object
- server should log deprecation message

trel: i like these three rules.

the twist is on the first one... we left `irm -U` documented/marked as deprecated since there was no other good place to put that notice to the user.   in this case, `adda` seemed out of place to begin with?  and it was documented in the running text above the subcommands, as an option.

happy to have it be more uniform, but also okay with it being a per-item decision about how to inform the human.

src/imeta.cpp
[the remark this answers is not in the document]
korydraughn: done.

src/imeta.cpp
korydraughn: i've captured the deprecation rule here: irods/irods_docs#124

alanking (approved):

korydraughn:
passed core tests.

added #

PR #275, opened 2022-02-23
[irods/irods#2919] Replace ireg -C with -r (main)

alanking:
CI tests are running

Companion to https://github.com/irods/irods/pull/6203

### DIFF
--- a/src/ireg.cpp
+++ b/src/ireg.cpp
@@ -1,15 +1,9 @@
-/*** Copyright (c), The Regents of the University of California            ***
- *** For more information please refer to files in the COPYRIGHT directory ***/
-/*
- * ireg - The irods reg utility
- */
-
-#include "rodsClient.h"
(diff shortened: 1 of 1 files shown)


src/ireg.cpp
```diff
@@ -28,7 +22,7 @@ main( int argc, char **argv ) {
     int nArgv;
 
 
-    optStr = "D:fhkKCG:R:vVZ";
+    optStr = "D:fhkKCrR:vVZ";
```
alanking, line 25: G is not processed in `parseCmdLineOpt` as it has to do with resource groups which were removed in 4.0.0. I figured it's about time to remove it. I can put it back if anybody disagrees

trel: torch approved.

alanking: Sweet

alanking:
Will merge along with irods/irods#6203

PR #262, opened 2022-01-22
[irods/irods5249] add iadmin set_delay_server subcommand (main)

trel:

PR #289, opened 2022-04-01
[irods/irods#5249] add iadmin get_delay_server_info subcommand (main)

korydraughn:
This PR builds upon the work of #262

### DIFF
--- a/src/iadmin.cpp
+++ b/src/iadmin.cpp
@@ -1,4 +1,5 @@
 #include <irods/filesystem.hpp>
+#include <irods/irods_at_scope_exit.hpp>
 #include <irods/irods_client_api_table.hpp>
 #include <irods/irods_pack_table.hpp>
 #include <irods/irods_string_tokenize.hpp>
@@ -6,6 +7,11 @@
 #include <irods/parseCommandLine.h>
 #include <irods/query_builder.hpp>
 #include <irods/rodsClient.h>
+#include <irods/rodsErrorTable.h>
+#include <irods/rodsLog.h>
+#include <irods/user_administration.hpp>
+#include <irods/user.hpp>
+#include <irods/get_grid_configuration_value.h>
 
 #include <algorithm>
 #include <iostream>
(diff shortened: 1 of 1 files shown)


src/iadmin.cpp
```diff
@@ -2387,6 +2460,24 @@ usage( char *subOpt ) {
...
 
+    char* get_delay_server_info_usage[] = {
+        "get_delay_server_info",
+        " ",
+        "Prints information about the delay server as JSON.",
+        " ",
+        "This command allows administrators to identify which server is running the",
+        "delay server and if the delay server is being migrated.",
+        " ",
+        "This information is retrieved from R_GRID_CONFIGURATION.",
```
trel, line 2471: "the R_GRID_CONFIGURATION table."

?

not sure it's necessary, just might be clarifying for a new reader.

src/iadmin.cpp
```diff
@@ -143,6 +149,68 @@ auto get_replica_value(
...
+            return CAT_INSUFFICIENT_PRIVILEGE_LEVEL;
+        }
+    }
+    catch (const irods::exception& e) {
+        std::cerr << "ERROR: " << e.client_display_what() << '\n'; // TODO Is there a common format for printing error messages?
+        return e.code();
+    }
+    catch (const std::exception& e) {
+        std::cerr << "ERROR: " << e.what() << '\n';
+        return 1; // TODO What error code should we use here?
```
trel, line 173: seems okay

src/iadmin.cpp
korydraughn: I actually had that originally lol, but tried to match set_delay_server's help text.

Will change to what you've recommended.

src/iadmin.cpp
korydraughn: Yeah, I guess this is correct. The OS doesn't know what the iRODS error code is. We can print the iRODS error code to the terminal and return a non-zero value for these cases.

src/iadmin.cpp
trel: nice - right back on me.     i'm sure i went with brevity at the time.   let's use more words.

korydraughn:
This PR is dependent upon https://github.com/irods/irods/pull/6272

Tests for this PR will be added following the merging of the irods/irods PR.

alanking (approved):

PR #290, opened 2022-04-04
[irods/irods_6276] Update userspace tarball packager

SwooshyCueb:
This PR does the following:
- [x] Updates the userspace packager's `pyproject.toml` to use `flakeheaven` instead of the abandoned `flakehell`
- [x] Fixes a bug in `utilbase.py` where `lief_ver_info` was being imported from the wrong place
- [x] Bumps the recommended version of LIEF to 0.12.
- [x] Adjusts warning threshold handling for slow ELF rebuilds as the ELF rebuilder in LIEF 0.12 is much faster than both 0.11 and 0.10's.
- [x] Marks LIEF's ELF rebuilder as fragile for versions prior to 0.12 and only performs segfault workarounds for fragile ELF rebuilders.
- Updates directive files for supported distros
    - [x] CentOS 7
    - [x] Ubuntu 18.04
- Adds directive files for newly supported distros
    - [x] Ubuntu 20.04
    - [x] Debian 11
    - [x] AlmaLinux 8
    - [x] Rocky Linux 8
- ~~Conditionally cleans out unneeded library imports with dangling symbol versions (lief-project/LIEF#670)~~ - later

Tested in containers:
- [x] CentOS 7
- [x] Ubuntu 18.04
- [x] Ubuntu 20.04
- [x] Debian 11
- [x] AlmaLinux 8
- [x] Rocky Linux 8

Closes irods/irods#6276

### DIFF
--- a/packaging/userspace/libdirectives.py
+++ b/packaging/userspace/libdirectives.py
@@ -205,6 +205,10 @@ class LibDirectives(PackagerUtilBase[LibDirectiveOptionsBase]):
 						soname = directive_l[1:]
 						self.allowlist.sonames.discard(soname)
 						self.denylist.sonames.add(soname)
+					elif directive_l[0] == '~':
+						soname = directive_l[1:]
+						self.allowlist.sonames.discard(soname)
+						self.denylist.sonames.discard(soname)
 					else:
 						l.warning('%s:%s: Syntax error, ignoring line', soname_dfname, str(lnum))
 		else:
--- a/packaging/userspace/libdirectives.py
+++ b/packaging/userspace/libdirectives.py
@@ -28,6 +28,10 @@ class LibDirectiveOptionsBase(PackagerOptionsBase):  # pylint: disable=W0223
 		'--target-platform', metavar='PLATFORM',
 		action='store', opt_type=str, default=None
 	)
+	target_platform_base: Optional[str] = RunArgOption(
+		'--target-platform-base', metavar='BASE',
+		action='store', opt_type=str, default=None
+	)
 	target_platform_variant: Optional[str] = RunArgOption(
 		'--target-platform-variant', metavar='VARIANT',
 		action='store', opt_type=str, default=None
@@ -161,18 +165,22 @@ class LibDirectives(PackagerUtilBase[LibDirectiveOptionsBase]):
 		self._denylist = DirectiveLibSets(context)
 
 		target_platform = context.options.target_platform
+		target_platform_base = context.options.target_platform_base
 		target_platform_variant = context.options.target_platform_variant
 		if target_platform == 'none':
 			target_platform = None
 		if target_platform_variant in ('none', 'rolling'):
 			target_platform_variant = None
+		if target_platform_base in ('none', ''):
+			target_platform_base = None
 		if target_platform and target_platform_variant:
 			target_platform = target_platform + target_platform_variant
 		self._target_platform = target_platform
+		self._target_platform_base = target_platform_base
 
-	def read_directives_for_target(self, target_name: str) -> NoReturn:
+	def read_directives_for_target(self, target_name: Optional[str]) -> bool:
 		if not target_name:
-			return
+			return True
 		l = logging.getLogger(__name__)  # noqa: VNE001,E741
 
 		libname_dfname = path.join(self.context.packager_dir, 'externs_libs.' + target_name)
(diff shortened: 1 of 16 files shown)


packaging/userspace/strip_util.py
```diff
@@ -204,7 +204,7 @@ def _clean_unused_dependencies(
 		lief = self.lief
 
 		# lief tends to segfault with libc++
-		if fname.startswith('libc++.') or fname.startswith('libc++abi.'):
+		if self.lief_info.fragile_builder and (fname.startswith('libc++.') or fname.startswith('libc++abi.')):
```
SwooshyCueb, line 207: missed one of these on `runpath_util.py:370`

alanking (approved):
Truthfully I'm not familiar with most of the changes, but what I've seen seems okay. Let's see if @korydraughn or @trel have thoughts

packaging/userspace/libdirectives.py
```diff
@@ -205,24 +213,42 @@ def read_directives_for_target(self, target_name: str) -> NoReturn:
 						soname = directive_l[1:]
 						self.allowlist.sonames.discard(soname)
 						self.denylist.sonames.add(soname)
+					elif directive_l[0] == '~':
+						soname = directive_l[1:]
+						self.allowlist.sonames.discard(soname)
+						self.allowlist.sonames.discard(soname)
```
trel, line 219: is twice... intentional?

packaging/userspace/libdirectives.py
```diff
@@ -205,24 +213,42 @@ def read_directives_for_target(self, target_name: str) -> NoReturn:
...
+		return False
 
 	def read_directives(self) -> NoReturn:
 		l = logging.getLogger(__name__)  # noqa: VNE001,E741
 		l.info('reading known libs data')
 
 		self.read_directives_for_target(target_name='common')
-		self.read_directives_for_target(target_name=self.target_platform)
+
+		# FIXME: Account for populated target_platform_base but empty target_platform
```
trel, line 231: still need to be handled?

packaging/userspace/userspace-packager.py
```diff
@@ -127,16 +130,26 @@ def __init__(
 		super().__init__(context, executor)
 
 		if not self.options.target_platform:
-			distro = linux_distribution(full_distribution_name=False)
-			self.options.target_platform = distro[0].lower()
-			self.options.target_platform_variant = distro[1].partition('.')[0].lower()
+			plat = distro.linux_distribution(full_distribution_name=False)
+			self.options.target_platform = plat[0].lower()
+			self.options.target_platform_variant = plat[1].partition('.')[0].lower()
+		# This will likely change once we are building on ubi8
```
trel, line 136: what does this mean?  ub18...

packaging/userspace/runpath_util.py
```diff
@@ -212,10 +212,13 @@ def _set_rpath_lief(
 				ent_runpath = lib.get(lief.ELF.DYNAMIC_TAGS.RUNPATH)
 				ent_runpath.paths = [runpath]
 			elif lib.has(lief.ELF.DYNAMIC_TAGS.RPATH):
-				# Changing the tag here causes a segfault when writing
 				ent_runpath = lib.get(lief.ELF.DYNAMIC_TAGS.RPATH)
 				ent_runpath.paths = [runpath]
-				runpath_is_rpath = True
+				if self.lief_info.fragile_builder:
+					# Changing the tag here causes a segfault when writing
```
korydraughn, line 218: Is this something to worry about? How does one get into this situation?

packaging/userspace/runpath_util.py
```diff
@@ -299,7 +302,7 @@ def _set_rpath_chrpath(
 				os.unlink(tmp_out)
 			op_state.chrpath.tool_tried = True
 
-	# FIXME: runpath setting with cmake is broken on centos
+	# FIXME: runpath setting with cmake fails silently on centos 7
```
korydraughn, line 305: Is this something to worry about?

korydraughn (approved):
I'm not familiar with everything this does, but like @alanking said, it appears solid. And given that your testing passed, I think this fine to merge.

Are there things we/users need to keep in mind when handling this?

Should docs.irods.org grow a section around this?

packaging/userspace/libdirectives.py
SwooshyCueb: Doesn't come up with the current directive files.

packaging/userspace/userspace-packager.py
SwooshyCueb: red hat universal base image. It's what rocky, alma, oracle, amazon, centos, et al are built on top of.

packaging/userspace/runpath_util.py
SwooshyCueb: No, the if block here works around it for problematic versions of lief

packaging/userspace/runpath_util.py
SwooshyCueb: Only if a an older version of lief is being used and `chrpath` is not present on the system.

packaging/userspace/libdirectives.py
SwooshyCueb: Nope, thanks for catching that

SwooshyCueb:
> Are there things we/users need to keep in mind when handling this?
>
> Should docs.irods.org grow a section around this?

Potentially. If there's documentation to be written, I'm not sure if I'm the right choice to handle that.

Here's some info that might help:
- I successfully built packages using docker images from my dev environment PR, and did my testing on vanilla (or mostly-vanilla in the case of `redhat/ubi8-minimal`) container images.
    - When running the builder images, I ran the `docker run` command almost exactly as `run_debugger.sh` does, only I passed `--entrypoint=""` so that I could pop a shell. From there, you can just run `/icommands_source/packaging/userspace/build_and_package.example.sh` This can probably be shortened to one step by passing `--entrypoint="/icommands_source/packaging/userspace/build_and_package.example.sh"` instead.
- Using older versions of `lief` is partially supported but may cause problems. Usually these are minor, but in the case of CentOS 7, an older version of lief could potentially result in bad `RUNPATH`s in the final package.
    - Outside of the example above, if something goes sideways, the packager is usually *very* noisy about it.
- For best results, build iRODS with a `CMAKE_INSTALL_PREFIX` set to something other than `/`, `/usr`, or `/usr/local`.
- The proper way to invoke the packager is via the `userspace-tarball` build target. The packager *can* be invoked directly, but requires an astronomical amount of command line arguments to do its job properly.
- `strip_util.py` and `runpath_util.py` can also be invoked directly.

packaging/userspace/userspace-packager.py
trel: oh, ubi, i totally read it as ubuntu18.   carry on.

SwooshyCueb:
\#'d

PR #292, opened 2022-04-20
Remove irm -U and -n (main)

alanking:
companion PR to irods/irods#6343

Tests running

### DIFF
--- a/src/irm.cpp
+++ b/src/irm.cpp
@@ -27,7 +27,7 @@ main( int argc, char **argv ) {
     rodsPathInp_t rodsPathInp;
 
 
-    optStr = "hfruUvVfn:Z"; // JMC - backport 4552
+    optStr = "hfruvVfn:Z"; // JMC - backport 4552
 
     status = parseCmdLineOpt( argc, argv, optStr, 1, &myRodsArgs ); // JMC - backport 4552
     if ( status < 0 ) {
@@ -37,12 +37,6 @@ main( int argc, char **argv ) {
     if ( myRodsArgs.help == True ) {
         usage();
         exit( 0 );
-    }
-    // Since 4.2.8
-    // This deprecation message is specific to irm, not to general unlinking
-    if (True == myRodsArgs.unmount) {
-        printf("-U is deprecated.  Please use iunreg instead.\n");
-        exit(1);
     }
 
     if ( argc - optind <= 0 ) {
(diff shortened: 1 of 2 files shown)


src/irm.cpp
```diff
@@ -110,35 +104,22 @@ usage() {
...
         "are unregistered and left on disk as-is.",
         " ",
-        "The -U option allows the unregistering of the data object or collection",
-        "without deleting the physical file. Normally, a normal user cannot",
-        "unregister a data object if the physical file is located in a resource",
-        "vault. The acNoChkFilePathPerm rule allows this check to be bypassed.",
-        " ",
         "There is no -R option (remove replica from a named resource) at this time.",
         "Please use itrim (with the -S option) instead.",
         " ",
```
korydraughn, line 114: This should be changed to say something like:
```bash
To remove a specific replica, use itrim.
```
We don't have to mention -R or anything. They can read the help text for itrim.

@trel Thoughts?

korydraughn (commented):
Looks good.

alanking:
Still some replica numbers being passed to unlink in a few places, so need to mark this as draft til those get sorted out.

src/irm.cpp
alanking: Can do

test/rules/rulemsiDataObjUnlink.r
```diff
@@ -2,15 +2,14 @@ myTestRule {
...
-#    replNum - the replica number to be removed
 #    forceFlag= - flag to remove file without transferring to trash
 #    irodsAdminRmTrash - flag to allow administrator to remove trash
 #    irodsRmTrash - flag for user to remove trash
 #Output parameter is:
 #  Status
 #Output from running the example is:
 #  Replica number 1 of file /tempZone/home/rods/sub1/objunlink2 is removed
-  msiDataObjUnlink("objPath=*SourceFile++++replNum=1",*Status);
+  msiDataObjUnlink("objPath=*SourceFile",*Status);
```
alanking: I'm not sure what this test is trying to accomplish, but we don't support `replNum` for this API anymore. I removed it, but this used to unlink the physical data. This call now moves both replicas to the trash. Is this acceptable? We don't even verify that anything happened in this test lol

src/irm.cpp
alanking: Done

test/rules/rulemsiDataObjUnlink.r
trel: yeah, fix the comment and they're both now in the trash.

these aren't as much tests as 'example rules' for how to hold these things.   so this is fine.

alanking: That's fair. I fixed the comments in the following commit so I think we're good. Thanks

alanking:
I believe this is ready for review now

src/irm.cpp
```diff
@@ -110,35 +104,23 @@ usage() {
...
         "Remove one or more data-object or collection from iRODS space. By default, ",
```
korydraughn, line 105: "data-object or collection" -> "data objects and/or collections"

I see use of `data-object` and `data object`.

Please update the help text so that only one form is used.

korydraughn (changes requested):

src/irm.cpp
alanking: Will do

alanking: Done. Also reworded a couple of the statements for clarification.

trel (approved):

alanking:
Squashed and rebased, will add #'s and merge on green in main PR (unless there are further comments)

korydraughn (approved):

PR #300, opened 2022-05-12
Fix native auth TTL and PAM prompt, remove options to show password in iinit (main)

alanking:
Companion PR: https://github.com/irods/irods/pull/6391

Still need to run through full test suite, but things were working when tested by hand

### DIFF
--- a/src/iinit.cpp
+++ b/src/iinit.cpp
@@ -22,7 +22,6 @@
 #include <fstream>
 
 void usage( char *prog );
-void usageTTL();
 
 /* Uncomment the line below if you want TTL to be required for all
    users; i.e. all credentials will be time-limited.  This is only
@@ -76,8 +75,7 @@ int main( int argc, char **argv )
 {
     signal( SIGPIPE, SIG_IGN );
 
-    int i = 0, ix = 0, status = 0;
-    int echoFlag = 0;
+    int i = 0, status = 0;
     rodsEnv my_env;
     rcComm_t *Conn = 0;
     rErrMsg_t errMsg;
@@ -85,19 +83,10 @@ int main( int argc, char **argv )
     int doPassword = 0;
     bool doingEnvFileUpdate = false;
 
-    status = parseCmdLineOpt( argc, argv, "ehvVlZ", 1, &myRodsArgs );
+    status = parseCmdLineOpt( argc, argv, "hvVlZ", 1, &myRodsArgs );
     if ( status != 0 ) {
         printf( "Use -h for help.\n" );
         return 1;
-    }
-
-    if ( myRodsArgs.echo == True ) {
-        echoFlag = 1;
-    }
-
-    if ( myRodsArgs.help == True && myRodsArgs.ttl == True ) {
-        usageTTL();
-        return 0;
     }
 
     if ( myRodsArgs.help == True ) {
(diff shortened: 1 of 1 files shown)


src/iinit.cpp
```diff
@@ -459,4 +435,12 @@ void usageTTL() {
     printf( "administrator (usually a few days).  With the --ttl option, you can\n" );
     printf( "specify how long this derived password will be valid, within the\n" );
     printf( "limits set by the administrator.\n" );
+    printf( "\n" );
+    printf( " -l  list the iRODS environment variables (only)\n" );
+    printf( " -v  verbose\n" );
+    printf( " -V  Very verbose\n" );
+    printf( "--ttl ttl  set the password Time To Live (specified in hours)\n" );
+    printf( "           Run 'iinit -h --ttl' for more\n" );
+    printf( " -h  this help\n" );
```
korydraughn, line 444: Let's add an `Options:` header just above L439.

Consider aligning the option descriptions. Maybe `--ttl ttl`'s text is moved to the line below and aligned with the others.

Also, `--ttl ttl` -> `--ttl TTL`

src/iinit.cpp
```diff
@@ -431,22 +421,8 @@ int main( int argc, char **argv )
...
     printf( "to be used automatically by the icommands.\n" );
-    printf( "Usage: %s [-ehvVl] [--ttl TimeToLive] [password]\n", prog );
-    printf( " -e  echo the password as you enter it (normally there is no echo)\n" );
-    printf( " -l  list the iRODS environment variables (only)\n" );
-    printf( " -v  verbose\n" );
-    printf( " -V  Very verbose\n" );
-    printf( "--ttl ttl  set the password Time To Live (specified in hours)\n" );
-    printf( "           Run 'iinit -h --ttl' for more\n" );
-    printf( " -h  this help\n" );
+    printf( "Usage: %s [-hvVl] [--ttl TimeToLive]\n", prog );
```
korydraughn, line 424: Let's capitalize the `--ttl` argument placeholder. That appears to be the convention followed by many Linux commands (e.g. `--ttl TTL`).

korydraughn (commented):
Code looks fine to me.

src/iinit.cpp
alanking: Sounds good

src/iinit.cpp
alanking: Will do

src/iinit.cpp
alanking: Following commit addresses it. If it helps, here's what it looks like:
```
iinit -h
Creates a file containing your iRODS password in a scrambled form,
to be used automatically by the icommands.
Usage: iinit [-hvVl] [--ttl TTL]

When using regular iRODS passwords you can use --ttl (Time To Live)
to request a credential (a temporary password) that will be valid
for only the number of hours you specify (up to a limit set by the
administrator).  This is more secure, as this temporary password
(not your permanent one) will be stored in the obfuscated
credential file (.irodsA) for use by the other iCommands.

When using PAM, iinit always generates a temporary iRODS password
for use by the other iCommands, using a time-limit set by the
administrator (usually a few days).  With the --ttl option, you can
specify how long this derived password will be valid, within the
limits set by the administrator.

Options:
 -l  list the iRODS environment variables (only)
 -v  verbose
 -V  Very verbose
 --ttl TTL
     set the password Time To Live (specified in hours)
 -h  this help

iRODS Version 4.3.0                iinit
```
Not sure if that's what you had in mind

trel:
looking good to me

src/iinit.cpp
korydraughn: Yes, that is correct. Just one last thing. Add a newline above the "Usage:" line and I think this PR will be ready.

korydraughn (approved):

PR #302, opened 2022-05-15
[irods/irods5277] update ichmod permission model help text (main)

trel:


src/ichmod.cpp
```diff
@@ -144,45 +144,57 @@ main( int argc, char **argv ) {
...
-        "Usage: ichmod [-rhvVM] null|read|write|own userOrGroup dataObj|Collection ...",
+        "Usage: ichmod [-rhvVM] AccessLevel userOrGroup dataObject|Collection ...",
         " or    ichmod [-rhvVM] inherit Collection ...",
         " or    ichmod [-rhvVM] noinherit Collection ...",
-        " -r  recursive - set the access level for all dataObjects",
-        "             in the entered collection and subCollections under it",
+        " ",
+        "Options:",
+        " -r  recursive - set the AccessLevel for all dataObjects",
+        "             in the entered Collection and subCollections under it",
```
korydraughn, line 153: Consider ...
```
-r  Recursively set the access level for all data objects in the
    collection.
```

src/ichmod.cpp
```diff
@@ -144,45 +144,57 @@ main( int argc, char **argv ) {
...
+        " - 'own'",
+        " - 'delete_object'",
+        " - 'write', 'modify_object'",
+        " - 'create_object'",
+        " - 'delete_metadata'",
+        " - 'modify_metadata'",
+        " - 'create_metadata'",
+        " - 'read', 'read_object'",
+        " - 'read_metadata'",
+        " - 'null'",
```
korydraughn, line 177: These will likely need to be reverted to the versions that do not use the underscores because of backwards compatibility.

src/ichmod.cpp
```diff
@@ -144,45 +144,57 @@ main( int argc, char **argv ) {
 void
 usage() {
     char *msgs[] = {
-        "Usage: ichmod [-rhvVM] null|read|write|own userOrGroup dataObj|Collection ...",
+        "Usage: ichmod [-rhvVM] AccessLevel userOrGroup dataObject|Collection ...",
```
korydraughn, line 147: We should change the placeholders to be in all caps and then use those throughout the help text, just like other Unix commands. We do that for `itouch` today.
- `AccessLevel` -> `ACCESS_LEVEL`
- `userOrGroup` -> `USER_OR_GROUP`
- `dataObject`/`Collection` -> `LOGICAL_PATH`

src/ichmod.cpp
trel: Discush.   Want this to be the good version....

src/ichmod.cpp
trel: ah, will do.

src/ichmod.cpp
trel:
```
        " -r  recursive - Recursively set the ACCESS_LEVEL for all DataObjects",
        "             in the named Collection and in all subCollections",
```
 ?

korydraughn: I don't think having "recursive" after "-r" is helpful. The help text immediately following it describes the option just fine. Also, we can align the next line with the other option's help text.

So, perhaps something like ...
```bash
 -r  Recursively set the ACCESS_LEVEL for all data objects and
     collections in LOGICAL_PATH. LOGICAL_PATH must represent a
     collection.
```

trel: done

trel:
remember to add irods/irods#844 to this commit message too

src/ichmod.cpp
```diff
@@ -193,6 +193,12 @@ usage() {
         " ",
         "The original owner of a dataObject can still set an ACCESS_LEVEL, even",
         "if their ACCESS_LEVEL has been set to 'null'.",
+        " ",
+        "Example Operations:",
+        " - irm - requires 'delete_object' or greater",
+        " - imv - requires 'delete_object' or greater, due to removal of old name",
+        " - iput - requires 'modify_object' or greater",
+        " - iget - requires 'read_object' or greater",
```
trel, line 201: open to better suggestions on how to do/show this type of information

src/ichmod.cpp
trel: discussed - [it should behave](https://github.com/irods/irods/pull/6397#pullrequestreview-974230895)

src/ichmod.cpp
korydraughn: I understand why these are here, but they feel out of place.

It might be cleaner to document the permissions for an operation in that operation's help text. For example, `irm` grows a line or two explaining what permission is needed to delete an object. Same for `iput` and other icommands.

korydraughn (approved):
Overall, this PR is fine.

Feel free to ignore my comments if you want. We can always iterate on this after releasing 4.3.0. We also have to implement the new irods CLI.

src/ichmod.cpp
```diff
@@ -144,49 +138,67 @@ main( int argc, char **argv ) {
...
+        " -r  Recursively set the ACCESS_LEVEL for all DataObjects",
+        "     and Collections in LOGICAL_PATH.  LOGICAL_PATH must",
+        "     represent a Collection.",
         " -v  verbose",
         " -V  Very verbose",
         " -M  Admin Mode",
         " -h  this help",
         " ",
-        "Modify access to dataObjects (iRODS files) and collections (directories).",
+        "Modify access to dataObjects and Collections.",
```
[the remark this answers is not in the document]
trel: Understood/agreed on the casing - I was attempting consistency with what we've had for a while.

Your later comment below suggests to leave it for now - agreed - let's come back to this across the board later.

trel:
Will leave for now - agreed - iterate later here and in new CLI.

src/ichmod.cpp
trel: I had written cool words here - but they got eaten.    Agreed on the weirdness/casing, was aiming for consistency with what we've had for a while.

trel: Oh, and I typed in the wrong box.    Yes, maybe these should be in different places... but since `ichmod` is where the permissions get changed... this also made some sense.   Maybe if the list were a little longer?  Some metadata levels too?   Or, not mentioning them related to the iCommands, but rather... just 'remove' and 'rename' and 'put' and 'get'?

alanking:
This PR should also handle a new build error in `itree` caused by the new filesystem permission enumerated values. I have the fixes in a branch here: https://github.com/irods/irods_client_icommands/compare/main...alanking:5277.alan

Tests running now

trel:
updated, incorporated alan's `itree` tweaks

alanking (approved):
Confirmed that tests are passing. Let's squash this down

trel:
squashed to 3

alanking:
Looks good. Please #

trel:
#'d

PR #304, opened 2022-05-18
[irods/irods#4481] Updated help text for "iadmin suq" (main)

korydraughn:


trel:
seem good upon inspection.

korydraughn:
I avoided explained how `sgq` can be used to simulate user quotas because that is not the responsibility of the help text.

docs.irods.org is a better place for that.

trel:
right - thought through that - i think your edits are clean enough.   if the tests are passing - ship em.

alanking (approved):

korydraughn:
All core tests pass.

#'d

PR #309, opened 2022-07-07
[#6128][#6127] Prevent rodsadmins from downgrading each others' user type

phdavis1027:
#6128 #6127 Previously, rodsadmin users could downgrade the user types of other rodsadmins, including themselves. This PR checks on the client side whether the target user of an ```iadmin moduser``` command is themselves a rodsadmin and throws SYS_NOT_ALLOWED if an attempt is being made to downgrade them.

### DIFF
--- a/src/iadmin.cpp
+++ b/src/iadmin.cpp
@@ -46,6 +46,8 @@ int lastCommandStatus = 0;
 
 void usage( char *subOpt );
 
+namespace ixa = irods::experimental::administration;
+
 namespace {
 
 namespace fs = irods::experimental::filesystem;
@@ -242,7 +244,7 @@ auto modify_replica(
                     return 0 == attr.compare(a.first);
                 });
             const auto attr_in_genquery_attrs_and_not_in_denylist{
-                std::cend(genquery_attrs) != attr_pair && 
+                std::cend(genquery_attrs) != attr_pair &&
                 std::none_of(
                     std::cbegin(genquery_attrs_denylist),
                     std::cend(genquery_attrs_denylist),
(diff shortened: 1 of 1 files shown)


korydraughn:
iRODS supports many clients. However, this change only affects the icommands. We need to apply a change that solves the problem for all clients. That means the problem must be addressed from within the server.

trel:
What Kory said... but also, this is not just checking whether the target user is a rodsadmin - it's whether the target rodsadmin is the current user OR the target user is a rodsadmin that is running an iRODS server in this zone.

If the target user is a rodsadmin who is both different from the logged-in user AND not running an iRODS server, then it can / should be downgraded.

phdavis1027:
> What Kory said... but also, this is not just checking whether the target user is a rodsadmin - it's whether the target rodsadmin is the current user OR the target user is a rodsadmin that is running an iRODS server in this zone.
> 
> If the target user is a rodsadmin who is both different from the logged-in user AND not running an iRODS server, then it can / should be downgraded.

I think the "running an iRODS server" distinction was not clear to me. In a Zone Report for instance, the user running a server will be the username listed under ```zone_report[icat_server][service_account_environment][irods_user_name]``` ( or ```zone_report[servers][<some index to particular server>][service_account_environment][irods_user_name]```)?

trel:
Yes, I believe that is correct.  So build a set of usernames from that, plus the current user, and then check the set.

korydraughn:
Yes, `zone_report...[service_account_environment][irods_user_name]` should match the `irods_user_name` in /var/lib/irods/.irods/irods_environment.json.

PR #311, opened 2022-07-20
[irods/irods_docs#145] Fix ils replica status help text (main)

alanking:
- Add unused read-locked state to the list for completeness
- Correct the integer value associated with the write-locked state
- Clarifies statement about stale replicas


korydraughn:
This is basically the same text as what you just merged in irods_docs, right?

alanking:
Correct, it is the same text. The text originally appeared here, then made it over into the docs for Troubleshooting as well.

korydraughn (approved):

PR #324, opened 2022-10-06
[irods/irods#6576] Add pam_interactive to iinit (main)

alanking:
This change adds pam_interactive for consideration in the various twists and turns of logic in iinit. The changes mainly have to do with refactoring how iinit determines whether or not to show the password prompt.

---

Servicing irods/irods#6576

Based on work done by @stefan-wolfsheimer:
https://github.com/stefan-wolfsheimer/irods_client_icommands/commit/2c188dec8c7d0d40dad60bd5785fb30dcf7ae89c

Some other notes...
- I have only confirmed so far that this builds. Still needs to be tested
- `PAM_INTERACTIVE_SCHEME` would ideally be found as a string defined by a header in the `pam_interactive` auth plugin (like `native` and `pam_password` do)
- I am assuming that `pam_interactive` will not be available in the 4.2.x series

### DIFF
--- a/src/iinit.cpp
+++ b/src/iinit.cpp
@@ -8,7 +8,6 @@
 #include <irods/irods_pack_table.hpp>
 #include <irods/irods_pam_auth_object.hpp>
 #include <irods/parseCommandLine.h>
-#include <irods/plugins/auth/pam_password.hpp>
 #include <irods/rcConnect.h>
 #include <irods/rcMisc.h>
 #include <irods/rods.h>
@@ -19,9 +18,34 @@
 #include <fmt/format.h>
 #include <nlohmann/json.hpp>
 
+#include <algorithm>
 #include <filesystem>
 #include <fstream>
 #include <iostream>
+
+namespace
+{
+    namespace irods_auth = irods::experimental::auth;
+
+    constexpr const char* const AUTH_OPENID_SCHEME = "openid";
+    constexpr const char* const PAM_INTERACTIVE_SCHEME = "pam_interactive";
+    constexpr const char* const PAM_PASSWORD_SCHEME = "pam_password";
+
+    auto scheme_uses_iinit_password_prompt(const std::string_view _scheme) -> bool
+    {
+        const std::initializer_list<const std::string_view> no_password_prompt = {
+            AUTH_OPENID_SCHEME,
+            irods::AUTH_GSI_SCHEME,
+            irods::AUTH_PAM_SCHEME,
+            PAM_PASSWORD_SCHEME,
+            PAM_INTERACTIVE_SCHEME
+        };
+
+        return std::none_of(std::cbegin(no_password_prompt),
+                            std::cend(no_password_prompt),
+                            [&_scheme](const auto& _s) { return _scheme == _s; });
+    } // scheme_uses_iinit_password_prompt
+} // anonymous namespace
 
 void usage( char *prog );
 
@@ -35,7 +59,6 @@ void usage( char *prog );
 
 #define TTYBUF_LEN 100
 #define UPDATE_TEXT_LEN NAME_LEN*10
-const char *AUTH_OPENID_SCHEME = "openid";
 
 /*
  Attempt to make the ~/.irods directory in case it doesn't exist (may
@@ -85,8 +108,6 @@ printUpdateMsg() {
 
 int main( int argc, char **argv )
 {
-    namespace irods_auth = irods::experimental::auth;
-
     signal( SIGPIPE, SIG_IGN );
 
     int i = 0, status = 0;
@@ -94,7 +115,6 @@ int main( int argc, char **argv )
     rcComm_t *Conn = 0;
     rErrMsg_t errMsg;
     rodsArguments_t myRodsArgs;
-    int doPassword = 0;
     bool doingEnvFileUpdate = false;
 
     status = parseCmdLineOpt( argc, argv, "hvVlZ", 1, &myRodsArgs );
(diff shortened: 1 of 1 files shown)


alanking:
I have now tested this with the `pam_interactive` plugin (with some modifications) and it seems to work. Taking out of draft.

src/iinit.cpp
```diff
@@ -19,8 +19,32 @@
...
+    constexpr const char* const PAM_INTERACTIVE_SCHEME = "pam_interactive";
+
+    auto scheme_uses_iinit_password_prompt(const std::string_view _scheme) -> bool
+    {
+        const std::initializer_list<const std::string_view> no_password_prompt = {
+            AUTH_OPENID_SCHEME,
+            irods::AUTH_GSI_SCHEME,
+            irods::AUTH_PAM_SCHEME,
+            irods_auth::scheme::pam_password,
+            PAM_INTERACTIVE_SCHEME
```
trel, line 40: remind me again why we have two different flavors of magic string here?

can we make `irods_auth::scheme::pam_password` also... the same as the others?   or change all the others?

alanking: This is kind of a weird spot in which we find ourselves...

Historically, the iRODS core code (or, I guess in this case it would be the client libraries) has maintained the scheme names of the auth plugins. As you can see, some of these are actually maintained by `iinit` itself.

In the new auth plugin framework work, I thought it might be a good idea for each plugin to provide its own scheme name, which is why `pam_password` appears the way that it is. Looking back on this, maybe this is not such a good idea because any external plugins which may or may not be installed cannot be supported like this.

That being said, I think this problem is specific to `iinit` and the plugins will not need to provide their own scheme names in our next authentication client.

So... I'm willing to drop the `irods_auth::scheme::pam_password` in favor of a local string if that seems better.

trel: I defer.   Consistency is good - but also, this is all going to change before too long...

trel: This is definitely the grown-up solution.    Now... how do we get there gracefully...

alanking: I've created https://github.com/irods/irods/issues/6752 to investigate this .

For the near term, I think that a file-scoped constant would be good to replace this silly thing. :)

alanking:
Force-push involved a rebase. Nothing changed in the first commit

src/iinit.cpp
korydraughn: > For the near term, I think that a file-scoped constant would be good to replace this silly thing. :)

I agree.

korydraughn:
Is there anything blocking this?
Is it safe to merge?
Has it been tested?

alanking:
> Is there anything blocking this?

No

> Is it safe to merge?

Probably

> Has it been tested?

I... can't remember. Probably not. Trying to do that now.

alanking:
Okay, I have confirmed that this is working. It's the most minimal setup I could come up with:
```bash
# as iRODS service account
$ iadmin mkuser alice rodsuser
$ cat /etc/pam.d/irods
auth    required                        pam_permit.so
```
```bash
# as some Linux user
$ cat ~/.irods/irods_environment.json
{
    "irods_authentication_scheme": "pam_interactive"
}
$ iinit
One or more fields in your iRODS environment file (irods_environment.json) are
missing; please enter them.
Enter the host name (DNS) of the server to connect to: localhost
Enter the port number: 1247
Enter your irods user name: alice
Enter your irods zone: tempZone
Those values will be added to your environment file (for use by
other iCommands) if the login succeeds.

login: alice
$ ils
/tempZone/home/alice:
```
A couple of notes:
1. Didn't have to make a Linux user named "alice" for this to work
2. "alice" the rodsuser doesn't need a password
3. If you have SSL enabled, this plugin causes `iinit` to emit an error message:
```
remote addresses: 192.168.224.3 ERROR: [-]      /irods_source/lib/core/src/rcConnect.cpp:279:int rcDisconnect(rcComm_t *) :  status [SSL_SHUTDOWN_ERROR]  errno [] -- message [failed to call 'client stop']
        [-]     /irods_source/lib/core/src/sockComm.cpp:98:irods::error sockClientStop(irods::network_object_ptr, rodsEnv *) :  status [SSL_SHUTDOWN_ERROR]  errno [] -- message [failed to call 'client stop']
                [-]     /irods_source/plugins/network/src/ssl.cpp:593:irods::error ssl_client_stop(irods::plugin_context &, rodsEnv *) :  status [SSL_SHUTDOWN_ERROR]  errno [] -- message [error shutting down the SSL connection | error:1408F119:SSL routines:ssl3_get_record:decryption failed or bad record mac]
```
I think this is a bug in the plugin, not `iinit`.

alanking:
The demonstration above just confirms that the `iinit` password prompt is not forcibly shown, but only the stuff that `pam_interactive` needs to show to get through the authentication process.

korydraughn:
> 1. Didn't have to make a Linux user named "alice" for this to work

Why don't you need a Linux user?

> I think this is a bug in the plugin, not `iinit`.

Seems that way. If a bug appears, we can always address it leading up to the release.

alanking:
I think the only reason the `pam_permit` module even asks for a user/login is so that applications using the module don't break (https://man7.org/linux/man-pages/man8/pam_permit.8.html). As near as I can tell, you should only use the module for testing purposes, but maybe it has some utility beyond that which I don't know about.

Anyways, the point is that there doesn't need to be a Linux user because this module doesn't even check - it just returns success. I pointed it out because this particular flow was only achievable with the `pam_interactive` plugin (not `pam_password`, which is hard-coded to authenticate with a password for an existing Linux user).

korydraughn:
Got it.

Ultimately, even if we merge this, admins won't run the risk of breaking anything (in case there's a bug) unless they **opt** into using `pam_interactive`, correct?

alanking:
Right, you have to set the authentication scheme to pam_interactive and the pam_interactive plugin would need to be installed both in the client and on the server.

korydraughn:
In that case, if you feel this has been tested and you're happy with it, then let's get it merged.

alanking:
Okey dokey. I'm happy with it. Mergin

PR #334, opened 2022-11-11
[irods/irods#6725] icommands now compile using the same C++ standard as the server (main)

korydraughn:
Required to compile icommands and irods/irods#6718 with C++20 std.


alanking (approved):

korydraughn:
#'d

PR #336, opened 2022-12-02
[irods/irods_5411] Allow longer `HOME` directory

MartinFlores751:
Enable `iinit` to handle longer `$HOME` directory.

### DIFF
--- a/src/iinit.cpp
+++ b/src/iinit.cpp
@@ -4,7 +4,6 @@
 #include <irods/irods_environment_properties.hpp>
 #include <irods/irods_gsi_object.hpp>
 #include <irods/irods_kvp_string_parser.hpp>
-#include <irods/irods_kvp_string_parser.hpp>
 #include <irods/irods_native_auth_object.hpp>
 #include <irods/irods_pack_table.hpp>
 #include <irods/irods_pam_auth_object.hpp>
@@ -17,10 +16,12 @@
 
 #include <boost/lexical_cast.hpp>
 
+#include <fmt/format.h>
 #include <nlohmann/json.hpp>
 
-#include <iostream>
+#include <filesystem>
 #include <fstream>
+#include <iostream>
 
 void usage( char *prog );
 
@@ -43,27 +44,37 @@ const char *AUTH_OPENID_SCHEME = "openid";
  */
 int
 mkrodsdir() {
-    char dirName[NAME_LEN];
-    int mode;
-    char *getVar;
-#ifdef windows_platform
-    getVar = iRODSNt_gethome();
-#else
-    getVar = getenv( "HOME" );
-#endif
-    rstrcpy( dirName, getVar, NAME_LEN );
-    rstrcat( dirName, "/.irods", NAME_LEN );
-    mode = 0700;
-#ifdef _WIN32
-    iRODSNt_mkdir( dirName, mode );
-#else
-    int error_code = mkdir( dirName, mode );
-    int errsv = errno;
-    if ( error_code != 0 && errsv != EEXIST ) {
-        rodsLog( LOG_NOTICE, "mkdir failed in mkrodsdir with error code %d", error_code );
+    namespace fs = std::filesystem;
+
+    const char* home_dir{std::getenv("HOME")};
+    if (home_dir == nullptr) {
+        fmt::print(stderr, "environment variable HOME not set\n");
+        return -1;
     }
-#endif
-    return 0; /* no error messages as it normally fails */
+
+    fs::path irods_dir{home_dir};
+    irods_dir /= ".irods";
+
+    std::error_code err_code;
+    fs::create_directory(irods_dir, err_code);
+    if (err_code) {
+        fmt::print(
+            stderr,
+            "failed to create directory [{}] with the following error: [{}]\n",
+            irods_dir.string(), err_code.message());
+        return -1;
+    }
+
+    fs::permissions(irods_dir, fs::perms::owner_all, err_code);
+    if (err_code) {
+        fmt::print(stderr,
+                   "setting permissions for directory [{}] failed with the "
+                   "following error: [{}]\n",
+                   irods_dir.string(), err_code.message());
+        return -1;
+    }
+
+    return 0;
 }
 
 void
@@ -123,7 +134,9 @@ int main( int argc, char **argv )
     }
 
     // Create ~/.irods/ if it does not exist
-    mkrodsdir();
+    if (mkrodsdir() != 0) {
+        return 1;
+    }
 
     using json = nlohmann::json;
 
@@ -242,7 +255,7 @@ int main( int argc, char **argv )
         }
 
         if ( i != 0 ) {
-            rodsLogError( LOG_ERROR, i, "Save Password failure" );
+            rodsLogError( LOG_ERROR, i, "Failed to save password." );
             return 1;
         }
     }


src/iinit.cpp
```diff
@@ -43,27 +44,30 @@ const char *AUTH_OPENID_SCHEME = "openid";
...
+    namespace fs = std::filesystem;
+
+    fs::path irods_dir{std::getenv("HOME")};
+    irods_dir / ".irods";
+
+    std::error_code dir_ec;
+    fs::create_directory(irods_dir, dir_ec);
+    if (dir_ec) {
+      std::cerr
+          << "create_directory failed in mkrodsdir with the following error: ["
```
trel, line 56: gah, that 'mkrodsdir' naming convention...

can we rename that to 'dotirods' or '.irods/' or something less inside baseball...

src/iinit.cpp
```diff
@@ -43,27 +44,30 @@ const char *AUTH_OPENID_SCHEME = "openid";
...
-#ifdef _WIN32
-    iRODSNt_mkdir( dirName, mode );
-#else
-    int error_code = mkdir( dirName, mode );
-    int errsv = errno;
-    if ( error_code != 0 && errsv != EEXIST ) {
-        rodsLog( LOG_NOTICE, "mkdir failed in mkrodsdir with error code %d", error_code );
+    namespace fs = std::filesystem;
+
+    fs::path irods_dir{std::getenv("HOME")};
```
korydraughn, line 49: Need to check the return value of `std::getenv`.

We need to guard against weird environments.

src/iinit.cpp
```diff
@@ -43,27 +44,30 @@ const char *AUTH_OPENID_SCHEME = "openid";
...
+    irods_dir / ".irods";
```
korydraughn, line 50: `/` -> `/=`

src/iinit.cpp
```diff
@@ -43,27 +44,30 @@ const char *AUTH_OPENID_SCHEME = "openid";
...
+
+    std::error_code dir_ec;
+    fs::create_directory(irods_dir, dir_ec);
+    if (dir_ec) {
+      std::cerr
+          << "create_directory failed in mkrodsdir with the following error: ["
+          << dir_ec.message() << "]\n";
+      return -1;
     }
```
korydraughn, line 59: What happens if the directory already exists?

src/iinit.cpp
```diff
@@ -43,27 +44,30 @@ const char *AUTH_OPENID_SCHEME = "openid";
...
+
+    fs::path irods_dir{std::getenv("HOME")};
+    irods_dir / ".irods";
+
+    std::error_code dir_ec;
+    fs::create_directory(irods_dir, dir_ec);
+    if (dir_ec) {
+      std::cerr
+          << "create_directory failed in mkrodsdir with the following error: ["
+          << dir_ec.message() << "]\n";
```
korydraughn, line 57: Prefer using `fmt::print(stderr, ...)` instead of `std::cerr`.

src/iinit.cpp
```diff
@@ -43,27 +44,30 @@ const char *AUTH_OPENID_SCHEME = "openid";
...
+      return -1;
     }
-#endif
-    return 0; /* no error messages as it normally fails */
+
+    std::error_code perm_ec;
+    fs::permissions(irods_dir, fs::perms::owner_all, perm_ec);
```
korydraughn, line 62: You could reuse the error object from the previous call.

src/iinit.cpp
```diff
@@ -43,27 +44,30 @@ const char *AUTH_OPENID_SCHEME = "openid";
...
+    if (perm_ec) {
+      std::cerr << "setting permissions for directory [" << irods_dir
+                << "] failed in mkrodsdir with the following error: ["
+                << perm_ec << "]\n";
```
korydraughn, line 66: Prefer `fmt::print(stderr, ...)`

korydraughn (commented):
A couple things to take care of and I think we're good.

Also, the indentation needs to be adjusted to 4 spaces.

src/iinit.cpp
MartinFlores751: Double-checked [cppreference for `create_directory(...)`](https://en.cppreference.com/w/cpp/filesystem/create_directory) and it should not report an error if the directory exists.

> 1) ... If the function fails because p resolves to an existing directory, no error is reported. Otherwise on failure an error is reported.

trel:
"formating" -> "formatting" in the commit message

alanking (commented):
Up to you if you wanna address the comment. I think this looks real good

src/iinit.cpp
```diff
@@ -43,27 +44,37 @@ const char *AUTH_OPENID_SCHEME = "openid";
...
-#else
-    int error_code = mkdir( dirName, mode );
-    int errsv = errno;
-    if ( error_code != 0 && errsv != EEXIST ) {
-        rodsLog( LOG_NOTICE, "mkdir failed in mkrodsdir with error code %d", error_code );
+    namespace fs = std::filesystem;
+
+    const char* home_dir{std::getenv("HOME")};
+    if (home_dir == nullptr) {
+        fmt::print(stderr, "environment variable HOME not set!");
```
korydraughn, line 51: Missing `\n`

korydraughn (commented):
Let's avoid using backticks in commit messages.

If others feel strongly about allowing them, let me know and we can discuss it.

trel:
> Let's avoid using backticks in commit messages.

yeah, i think i agree.

src/iinit.cpp
```diff
@@ -18,9 +17,11 @@
 #include <boost/lexical_cast.hpp>
 
 #include <nlohmann/json.hpp>
+#include <fmt/format.h>
```
alanking, line 20: Well, since we're still debating stuff, can you flip the order of 19 and 20 for more alphabetization? :)

korydraughn:
Do we have a test that goes with this?

MartinFlores751:
Test are at irods/irods#6787.

src/iinit.cpp
```diff
@@ -123,7 +134,9 @@ int main( int argc, char **argv )
     }
 
     // Create ~/.irods/ if it does not exist
-    mkrodsdir();
+    if (mkrodsdir() != 0) {
+        std::exit(EXIT_FAILURE);
```
alanking, line 138: Any reason to prefer this over just returning a code from `main`? Really asking.

alanking (commented):
Just had a question. I think this is more or less ready to be squashed (omit #s for now) unless anybody objects

src/iinit.cpp
MartinFlores751: Think I did a `std::exit(...)` since I did that originally within `mkrodsdir(...)`... We can change that to a `return` instead.

korydraughn: Yeah, seeing `return` in `main()` feels more intuitive.

korydraughn:
Squash everything down and leave off the pounds.

src/iinit.cpp
```diff
@@ -43,27 +44,37 @@ const char *AUTH_OPENID_SCHEME = "openid";
...
-#else
-    int error_code = mkdir( dirName, mode );
-    int errsv = errno;
-    if ( error_code != 0 && errsv != EEXIST ) {
-        rodsLog( LOG_NOTICE, "mkdir failed in mkrodsdir with error code %d", error_code );
+    namespace fs = std::filesystem;
+
+    const char* home_dir{std::getenv("HOME")};
+    if (home_dir == nullptr) {
+        fmt::print(stderr, "environment variable HOME not set!\n");
```
trel, line 51: Find myself hesitating on the exclamation point.

In general, I don't think we've been too excited in the logs and in error messages to the user.

src/iinit.cpp
```diff
@@ -242,7 +255,7 @@ int main( int argc, char **argv )
         }
 
         if ( i != 0 ) {
-            rodsLogError( LOG_ERROR, i, "Save Password failure" );
+            rodsLogError( LOG_ERROR, i, "Failed to save password!" );
```
trel, line 258: same here.

MartinFlores751: I'll change it to a `.` here since the status code gets printed on the same line right after the message. It gets a bit confusing with no punctuation in my eyes.

Original message below for example:
`ERROR: Save Password failure status = -900000 FILE_OPEN_ERR`

trel: period is fine fine.  thanks.

trel:
i'm good for squashing.

alanking (commented):
Let's mention `iinit` in the commit message somewhere and I think this is good for #s

PR #338, opened 2022-12-12
[irods/irods#6398] iinit: Add options for auth scheme and SSL (main)

alanking:
Addresses irods/irods#6398

Yielding to #336 due to inevitable merge conflicts

Still need to run tests. Still a ways to go here, but putting this up here so we can have something to point at.

Some issues/discussion points:
 - Option names
 - Lots of repeated code when dealing with the integer-based settings (encryption)
 - Some workarounds/shims I'm not happy with (e.g. `_reloadRodsEnv`, forced `--always-save-environment` with `--with-ssl`)
 - Should we add other options? `--disable-ssl`?
 - Should we provide default values for the existing prompts?

### DIFF
--- a/src/iinit.cpp
+++ b/src/iinit.cpp
@@ -1,6 +1,7 @@
 #include <irods/authentication_plugin_framework.hpp>
 #include <irods/irods_auth_constants.hpp>
 #include <irods/irods_client_api_table.hpp>
+#include <irods/irods_configuration_keywords.hpp>
 #include <irods/irods_environment_properties.hpp>
 #include <irods/irods_gsi_object.hpp>
 #include <irods/irods_kvp_string_parser.hpp>
@@ -19,33 +20,10 @@
 #include <nlohmann/json.hpp>
 
 #include <algorithm>
+#include <iostream>
 #include <filesystem>
 #include <fstream>
 #include <iostream>
-
-namespace
-{
-    namespace irods_auth = irods::experimental::auth;
-
-    constexpr const char* const AUTH_OPENID_SCHEME = "openid";
-    constexpr const char* const PAM_INTERACTIVE_SCHEME = "pam_interactive";
-    constexpr const char* const PAM_PASSWORD_SCHEME = "pam_password";
-
-    auto scheme_uses_iinit_password_prompt(const std::string_view _scheme) -> bool
-    {
-        const std::initializer_list<const std::string_view> no_password_prompt = {
-            AUTH_OPENID_SCHEME,
-            irods::AUTH_GSI_SCHEME,
-            irods::AUTH_PAM_SCHEME,
-            PAM_PASSWORD_SCHEME,
-            PAM_INTERACTIVE_SCHEME
-        };
-
-        return std::none_of(std::cbegin(no_password_prompt),
-                            std::cend(no_password_prompt),
-                            [&_scheme](const auto& _s) { return _scheme == _s; });
-    } // scheme_uses_iinit_password_prompt
-} // anonymous namespace
 
 void usage( char *prog );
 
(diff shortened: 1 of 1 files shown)


alanking:
From some discussion, we have decided on the following...
 - The default behavior should be to always save the client environment file, so we don't need that option anymore (
 - We *should* provide defaults from the environment file for all prompts
 - Remove "Use default SSL encryption settings" in favor of default prompts

alanking:
Tests are all passing. Squashed to one and taking out of draft.

src/iinit.cpp
```diff
@@ -66,11 +69,225 @@ mkrodsdir() {
...
+
+        // The else case indicates that the response was empty and the environment already has a value. If the
+        // environment has a value already, the default value is the existing value (meaning, there will be no
+        // change in the environment). Therefore, no copy occurs in the else case as no change will occur.
+        if (!response.empty()) {
+            std::strncpy(_setting, response.c_str(), _len);
+        }
+        else if (!env_has_value) {
+            std::strncpy(_setting, default_value, _len);
+        }
```
trel, line 166: I feel the comment doesn't quite reflect the code...  please read again and confirm/deny.

src/iinit.cpp
```diff
@@ -66,11 +69,225 @@ mkrodsdir() {
...
+    auto set_env_from_prompt(int& _setting, int _default, const char* _prompt) -> void
+    {
+        const bool env_has_value = 0 != _setting;
+        const auto default_value = env_has_value ? _setting : _default;
+
+        fmt::print(fmt::runtime(fmt::format("{} [{{}}]: ", _prompt)), default_value);
+        std::string response;
+        std::getline(std::cin, response);
+
+        // The else case indicates that the response was empty and the environment already has a value. If the
```
trel, line 178: same.  not quite sure this is clear.

src/iinit.cpp
```diff
@@ -66,11 +69,225 @@ mkrodsdir() {
...
+        if (0 == std::strlen(_env.rodsZone)) {
+            constexpr const char* zone_prompt = "Enter your iRODS zone";
+            set_env_from_prompt(_env.rodsZone, zone_prompt, sizeof(_env.rodsZone));
+            _json_env[irods::KW_CFG_IRODS_ZONE] = _env.rodsZone;
+        }
+    } // configure_required_settings_in_env
+
+    auto configure_ssl_in_env(RodsEnvironment& _env, nlohmann::json& _json_env) -> void
+    {
+        // If the user indicated that SSL is going to be used, these settings are required, so no prompts occur.
```
trel, line 224: "these settings are required, so no prompts occur." -> "this setting is required, so no prompt shown."

src/iinit.cpp
```diff
@@ -66,11 +69,225 @@ mkrodsdir() {
...
+        constexpr const char* default_client_server_policy = "CS_NEG_REQUIRE";
+        std::strncpy(_env.rodsClientServerPolicy, default_client_server_policy, sizeof(_env.rodsClientServerPolicy));
+        _json_env[irods::KW_CFG_IRODS_CLIENT_SERVER_POLICY] = _env.rodsClientServerPolicy;
+
+        constexpr const char* default_server_negotiation = "request_server_negotiation";
```
trel, line 229: add same comment as above about no prompt.

src/iinit.cpp
```diff
@@ -66,11 +69,225 @@ mkrodsdir() {
...
+        std::strncpy(
+            _env.rodsClientServerNegotiation, default_server_negotiation, sizeof(_env.rodsClientServerNegotiation));
+        _json_env[irods::KW_CFG_IRODS_CLIENT_SERVER_NEGOTIATION] = _env.rodsClientServerNegotiation;
+
+        constexpr const char* default_server_verification = "cert";
+        constexpr const char* server_verification_prompt = "Enter the server verification level";
```
trel, line 235: consider showing a bracketed set of choices... the enum...
"Enter the server verification level [none, cert, hostname]"

also, consider that we say in the docs that 'hostname' is the default if it's not set... so maybe we should default to hostname here too?

src/iinit.cpp
```diff
@@ -84,7 +301,11 @@ int main( int argc, char **argv )
     rErrMsg_t errMsg;
     rodsArguments_t myRodsArgs;
     int doPassword = 0;
-    bool doingEnvFileUpdate = false;
+
+    // THESE MUST BE DONE HERE! parseCmdLineOpt is EVIL and considers any unknown options invalid.
```
trel, line 305: maybe this comment should move down a couple lines?

src/iinit.cpp
```diff
@@ -433,6 +567,8 @@ void usage( char *prog ) {
     printf( "The client environment file is sought in ~/.irods/irods_environment.json\n" );
```
trel, line 567: sought... weird.   consider rewording this sentence for non-native english speakers.

src/iinit.cpp
```diff
@@ -467,6 +603,11 @@ void usage( char *prog ) {
     printf( " -V  Very verbose\n" );
     printf( " --ttl TTL\n" );
     printf( "     set the password Time To Live (specified in hours)\n" );
+    printf(" --with-ssl\n");
+    printf("     Includes prompts which will set up SSL communications in the\n");
```
trel, line 607: singular feels more correct?

"Includes" -> "Include"

src/iinit.cpp
```diff
@@ -467,6 +603,11 @@ void usage( char *prog ) {
...
+    printf("     client environment.\n");
+    printf(" --prompt-auth-scheme\n");
+    printf("     Includes a prompt to select the authentication scheme.\n");
```
trel, line 610: singular feels more correct?

"Includes" -> "Include"

src/iinit.cpp
trel: "prompt shown" -> "prompt is shown"

src/iinit.cpp
```diff
@@ -66,11 +69,225 @@ mkrodsdir() {
...
+        else {
+            obj_to_dump = _update;
+        }
+
+        if (std::ofstream f(env_file.c_str(), std::ios::out); f.is_open()) {
+            f << obj_to_dump.dump(4) << "\n";
+            f.close();
+        }
+        else {
+            fmt::print("failed to open environment file [{}]\n", env_file);
```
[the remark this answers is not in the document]

src/iinit.cpp
```diff
@@ -66,11 +69,225 @@ mkrodsdir() {
...
+
+    auto option_specified(std::string_view _option, int argc, char** argv) -> bool
+    {
+        for (int arg = 0; arg < argc; ++arg) {
+            if (!argv[arg]) {
+                continue;
+            }
+
+            if (_option == argv[arg]) {
+                // parseCmdLineOpt is EVIL and requires this. Please don't ask why.
```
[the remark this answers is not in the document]
alanking: I am of the mind that we should save such improvements for the new CLI, but I'm not strongly opinionated on that.

src/iinit.cpp
alanking: The comment is explaining why there is no else case here (it's a no-op). Perhaps it would be more clear if the comment (or a separate comment) appeared where the else case would be if there was one? Open to suggestions on alternative wording as well.

src/iinit.cpp
```diff
@@ -66,11 +69,225 @@ mkrodsdir() {
...
+
+        // The else case indicates that the response was empty and the environment already has a value. If the
+        // environment has a value already, the default value is the existing value (meaning, there will be no
+        // change in the environment). Therefore, the else case is a no-op as no change will occur.
+        if (!response.empty()) {
+            try {
+                _setting = boost::lexical_cast<int>(response);
+            }
+            catch (const boost::bad_lexical_cast&) {
+                _setting = default_value;
```
alanking, line 186: Might consider printing an error message here, too.

src/iinit.cpp

src/iinit.cpp
alanking: Showing the choices makes me wonder why we wouldn't show choices (where applicable) for other questions (auth scheme, encryption algorithm, etc.). Plus, the behavior here is to use what's in the brackets as the default value if nothing is entered. How should we indicate the default value if we show the options (if indeed there is still a default)?

Didn't notice that "hostname" is the default. Will change to that once we decide on the above

src/iinit.cpp
alanking: No, this comment applies to lines 307 and 308. If these options remain in `argv` when `parseCmdLineOpt` is called, an error will occur and `iinit` will terminate.

src/iinit.cpp
alanking: "sought" -> "found"?

src/iinit.cpp

src/iinit.cpp

src/iinit.cpp
alanking: Oops, this didn't appear in the Files Changed tab.

I think the comment reflects the code. There is no else case and it is explaining what the else case would mean if there was code here to represent it. There is no code to represent it because it's a no-op.

src/iinit.cpp
alanking: Done

src/iinit.cpp
```diff
@@ -467,6 +604,11 @@ void usage( char *prog ) {
     printf( " -V  Very verbose\n" );
     printf( " --ttl TTL\n" );
     printf( "     set the password Time To Live (specified in hours)\n" );
+    printf(" --with-ssl\n");
+    printf("     Include prompts which will set up SSL communications in the\n");
+    printf("     client environment.\n");
```
korydraughn, line 609: Indentation does not match L606

src/iinit.cpp
```diff
@@ -467,6 +604,11 @@ void usage( char *prog ) {
...
+    printf(" --prompt-auth-scheme\n");
+    printf("     Include a prompt to select the authentication scheme.\n");
```
korydraughn, line 611: Indentation does not match L606

src/iinit.cpp
```diff
@@ -66,11 +69,226 @@ mkrodsdir() {
...
+                    e.what());
+            }
+
+            obj_to_dump.merge_patch(_update);
+        }
+        else {
+            obj_to_dump = _update;
+        }
+
+        if (std::ofstream f(env_file.c_str(), std::ios::out); f.is_open()) {
```
korydraughn, line 103: This line is slightly different than L84.

I think L84 is slightly better, mostly because of how the stream is checked.

src/iinit.cpp
```diff
@@ -66,11 +69,226 @@ mkrodsdir() {
...
+            f << obj_to_dump.dump(4) << "\n";
+            f.close();
```
korydraughn, line 105: This line could be removed.

Is there a reason for explicitly calling `f.close()`?

src/iinit.cpp
```diff
@@ -66,11 +69,226 @@ mkrodsdir() {
...
+            std::strncpy(_setting, response.c_str(), _len);
+        }
+    } // set_env_from_prompt
+
+    auto set_env_from_prompt(char _setting[], const char* _default, const char* _prompt, std::size_t _len) -> void
+    {
+        const bool env_has_value = 0 != std::strlen(_setting);
+        const char* default_value = env_has_value ? _setting : _default;
+
+        fmt::print(fmt::runtime(fmt::format("{} [{{}}]: ", _prompt)), default_value);
```
korydraughn, line 154: Seems this line could be simplified?
```cpp
fmt::print("{} [{}]: ", _prompt, default_value);
```

src/iinit.cpp
```diff
@@ -66,11 +69,226 @@ mkrodsdir() {
...
+    } // set_env_from_prompt
+
+    auto set_env_from_prompt(int& _setting, int _default, const char* _prompt) -> void
+    {
+        const bool env_has_value = 0 != _setting;
+        const auto default_value = env_has_value ? _setting : _default;
+
+        fmt::print(fmt::runtime(fmt::format("{} [{{}}]: ", _prompt)), default_value);
+        std::string response;
+        std::getline(std::cin, response);
```
korydraughn, line 176: Seems this line could be simplified?
```cpp
fmt::print("{} [{}]: ", _prompt, default_value);
```

src/iinit.cpp
korydraughn: It could be replaced with a table/list. Or, you can try inverting the explanation.

I don't feel it needs a comment. The code is small enough that it is pretty easy to understand.

src/iinit.cpp
alanking: Sure thing

src/iinit.cpp
alanking: Hmm... yes, I think you're right. This is probably a holdover from a previous iteration of this logic. Will update.

src/iinit.cpp
alanking: Yep

src/iinit.cpp
alanking: D'oh

src/iinit.cpp
alanking: Double d'oh

src/iinit.cpp
alanking: I believe it's because I put the `ofstream` declaration in the if statement and didn't update the logic accordingly. :) Will update.

src/iinit.cpp
alanking: Seeing as the comments have generated more confusion than they have alleviated, I think it may be better to just remove it. I find the code clear, personally, but that's because I wrote it. :)

src/iinit.cpp
alanking: Done

src/iinit.cpp
trel: maybe instead... just change the sentence to be declarative of the order of operations...

```
Environment information is loaded from the following locations, with priority being given to the top of the list:
 - in specific environment variables
 - in an irods_environment.json file located at IRODS_ENVIRONMENT_FILE
 - in ~/.irods/irods_environment.json
 - default values set in the server
 ```

also, we should probably test to see whether that is all true...
if there is an IRODS_ENVIRONMENT_FILE defined
AND
say... some specific things also set... IRODS_DEFAULT_RESOURCE...

do they behave / get honored in the way described above?

File: src/iinit.cpp [hunk 19]
```diff
@@ -433,6 +567,8 @@ void usage( char *prog ) {
     printf( "The client environment file is sought in ~/.irods/irods_environment.json\n" );
     printf( "(unless there is an environment variable indicating a different location),\n" );
     printf( "but the configurations can be set via environment variables as well.\n" );
+    printf( "The client environment file will be updated each time iinit is run in order\n" );
```
trel, line 563: "The client environment file" -> "The active client environment file"

trel:
#336 is now merged

src/iinit.cpp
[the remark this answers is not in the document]
alanking: I see there being three options here:
1. Leave the comment
2. Delete the comment
3. Actually answer the "why" question in the comment

I think the CLI will be better and so it's not worth an issue. Thoughts?

src/iinit.cpp
alanking: There are two ways we can go about this:
1. Show all choices available to the user in each prompt. I would think this should be considered a separate unit of work. We do not have strings for that stuff readily available in a library function for all of these prompts. Plus, we would need to consider what's available to the client versus the server, etc.
2. Show only the default choice to users when there is a default for a particular prompt, in which case nothing is shown (how it behaves now)

I think either option is fine, but I think if we go with option 1, we should make a new issue and move along.

As before, I'll fix this case so that it says "default".

src/iinit.cpp
alanking: I'll try this out and report back.

src/iinit.cpp
alanking: Will do

src/iinit.cpp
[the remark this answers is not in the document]
korydraughn: I don't have a strong preference regarding the comment. However, I don't think we should put time into investigating it or providing context around it.

When development on the new CLI takes off, no one will care about this.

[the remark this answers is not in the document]
alanking: In that case, I will leave the comment.

src/iinit.cpp
alanking: The default has been changed to "hostname".

I created https://github.com/irods/irods/issues/7142 to continue this discussion. For now, I'm going to just show the default values and we can change it before release. If we would rather show nothing, just open a new thread.

src/iinit.cpp
alanking: I'm not familiar with all of the environment variables, so I'll come back to testing that one. However, based on the code, for populating the client environment struct, it seems true that the environment variables take highest priority: https://github.com/irods/irods/blob/2979e35257c9e6454568371338a82862a3ea3718/lib/core/src/getRodsEnv.cpp#L121-L126 The environment file is consulted first, then indiscriminately overwrites values with those from environment variables (i.e. `getenv`) where they exist.

Confirmed that `IRODS_ENVIRONMENT_FILE` affects where `iinit` looks for the active client environment file (ignore the root-ness):
```bash
$ ls -l ~/.irods/
total 4
-rw-r--r-- 1 root root 169 Jun 26 18:34 irods_environment.json
$ IRODS_ENVIRONMENT_FILE=/lol.json iinit
 ERROR: environment_properties::capture: missing environment file. should be at [/lol.json]
Enter the host name (DNS) of the server to connect to: 
```
If not specified, defaults to looking in `~/.irods`, so that works as well.

src/iinit.cpp

src/iinit.cpp
alanking: I updated the help text to read just like @trel suggested.

Tested out the environment variable thing and it worked out as expected. For this demonstration, the default resource is "demoResc" in the server:
```bash
$ cat ~/.irods/irods_environment.json  # no default resource in client environment
{
    "irods_host": "localhost",
    "irods_port": 1247,
    "irods_user_name": "alan",
    "irods_zone_name": "tempZone"
}
$ echo $IRODS_DEFAULT_RESOURCE     

$ iput putme foo1
$ cat ~/.irods/irods_environment.json # add a default resource to the client environment
{
    "irods_default_resource": "ufs0",
    "irods_host": "localhost",
    "irods_port": 1247,
    "irods_user_name": "alan",
    "irods_zone_name": "tempZone"
}
$ iput putme foo2
$ IRODS_DEFAULT_RESOURCE=ufs1 iput putme foo3
$ ils -l
/tempZone/home/alan:
  alan              0 demoResc          418 2023-06-26.19:11 & foo1
  alan              0 ufs0          418 2023-06-26.19:12 & foo2
  alan              0 ufs1          418 2023-06-26.19:12 & foo3
```

alanking:
Force-push was due to rebase. Haven't squashed anything. Last two commits show an author date of today, so that's the new stuff.

src/iinit.cpp
korydraughn: Looks good to me.

alanking: Yay, resolving

alanking:
Before I forget...

Possibly supersedes #56

alanking:
Any other thoughts before I rebase/squash?

korydraughn (approved):
If you're happy with it then let's get it merged.

trel:
no thoughts!

PR #349, opened 2023-01-19
[irods#6740] shaping up iqstat (4-2-stable)

d-w-moore:
Fix the things that are broken about `iqstat` options, and maybe even  a little more generally.  "Touch, improve."
  - `iqstat -a <anything>` - caused agent crash which made the client report SYS_INTERNAL_ERR
    * now, `<anything>` is scanned and rejected if not formatted as unsigned int.
    * plus, `iqstat -a <unsigned_int_rule_id>` works now.  it reports all rule in the system regardless of the owning user
  - where there were two routines, the brief and long version, they are now integrated into one.
  - there may be more, i just can't remember right now. more later
  - if any of these things expose security problems, let's discuss.

### DIFF
--- a/src/iqstat.cpp
+++ b/src/iqstat.cpp
@@ -1,7 +1,19 @@
 #include "irods_client_api_table.hpp"
 #include "irods_pack_table.hpp"
+#include "irods_query.hpp"
+#include "rcMisc.h"
 #include "rods.h"
 #include "rodsClient.h"
+#include "rodsLog.h"
+#include "user_administration.hpp"
+
+#include <boost/algorithm/string/trim.hpp>
+#include <boost/format.hpp>
+
+#include <cstdint>
+#include <cstdio>
+#include <stdexcept>
+#include <string>
 
 #define MAX_SQL 300
 #define BIG_STR 200
@@ -10,21 +22,21 @@ char cwd[BIG_STR];
 
 int debug = 0;
 
-rcComm_t *Conn;
+rcComm_t *Conn{};
 rodsEnv myEnv;
 
 void usage();
 
+auto throw_if_id_cannot_be_converted_to_int(const std::string& key) -> void;
+
 /*
  print the results of a general query.
  */
-int
+void
 printGenQueryResults( rcComm_t *Conn, int status, genQueryOut_t *genQueryOut,
                       char *descriptions[], int formatFlag ) {
-    int printCount;
     int i, j;
     char localTime[TIME_LEN];
-    printCount = 0;
     if ( status != 0 ) {
         printError( Conn, status, "rcGenQuery" );
     }
@@ -56,7 +68,6 @@ printGenQueryResults( rcComm_t *Conn, int status, genQueryOut_t *genQueryOut,
                             printf( "%s ", tResult );
                         }
                     }
-                    printCount++;
                 }
                 if ( formatFlag == 1 ) {
                     printf( "\n" );
@@ -64,201 +75,57 @@ printGenQueryResults( rcComm_t *Conn, int status, genQueryOut_t *genQueryOut,
             }
         }
     }
-    return 0;
 }
 
 /*
 Via a general query, show rule information
 */
-int
-showRuleExec( char *name, char *ruleName, int allFlag ) {
-    genQueryInp_t genQueryInp;
-    genQueryOut_t *genQueryOut;
-    int i1a[20];
-    int i1b[20] = {0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0};
-    int i2a[20];
-    char *condVal[10];
-    char v1[BIG_STR];
-    char v2[BIG_STR];
-    int i, status;
-    int printCount;
-    char *columnNames[] = {"id", "name", "rei_file_path", "user_name",
-                           "address", "time", "frequency", "priority",
-                           "estimated_exe_time", "notification_addr",
-                           "last_exe_time", "exec_status"
-                          };
 
-    printCount = 0;
-    memset( &genQueryInp, 0, sizeof( genQueryInp_t ) );
+namespace
+{
+    /* This structure encapsulates a number of parameters having generally to do with
+       the output choices of "brief" or "long" formats when listing the information
+       for each rule ID; in a nutshell, which columns are listed, and how many.
+       The info given in a "brief" is a subset (of the first two) of the "long" listing.
+     */
+    struct output_format_parameters {
+        int columns[20]{};
+        int brief_format_len{};
+        int long_format_len{};
 
-    i = 0;
-    i1a[i++] = COL_RULE_EXEC_ID;
-    i1a[i++] = COL_RULE_EXEC_NAME;
-    i1a[i++] = COL_RULE_EXEC_REI_FILE_PATH;
-    i1a[i++] = COL_RULE_EXEC_USER_NAME;
-    i1a[i++] = COL_RULE_EXEC_ADDRESS;
-    i1a[i++] = COL_RULE_EXEC_TIME;
-    i1a[i++] = COL_RULE_EXEC_FREQUENCY;
-    i1a[i++] = COL_RULE_EXEC_PRIORITY;
-    i1a[i++] = COL_RULE_EXEC_ESTIMATED_EXE_TIME;
-    i1a[i++] = COL_RULE_EXEC_NOTIFICATION_ADDR;
-    i1a[i++] = COL_RULE_EXEC_LAST_EXE_TIME;
-    i1a[i++] = COL_RULE_EXEC_STATUS;
+        constexpr output_format_parameters() {
+            int i = 0;
+            columns[i++] = COL_RULE_EXEC_ID;
+            columns[i++] = COL_RULE_EXEC_NAME;
 
-    genQueryInp.selectInp.inx = i1a;
-    genQueryInp.selectInp.value = i1b;
-    genQueryInp.selectInp.len = i;
+            brief_format_len = i;
 
-    if ( allFlag ) {
-        genQueryInp.sqlCondInp.len = 0;
-    }
-    else {
-        i2a[0] = COL_RULE_EXEC_USER_NAME;
-        snprintf( v1, BIG_STR, "='%s'", name );
-        condVal[0] = v1;
-        genQueryInp.sqlCondInp.inx = i2a;
-        genQueryInp.sqlCondInp.value = condVal;
-        genQueryInp.sqlCondInp.len = 1;
-    }
+            columns[i++] = COL_RULE_EXEC_REI_FILE_PATH;
+            columns[i++] = COL_RULE_EXEC_USER_NAME;
+            columns[i++] = COL_RULE_EXEC_ADDRESS;
+            columns[i++] = COL_RULE_EXEC_TIME;
+            columns[i++] = COL_RULE_EXEC_FREQUENCY;
+            columns[i++] = COL_RULE_EXEC_PRIORITY;
+            columns[i++] = COL_RULE_EXEC_ESTIMATED_EXE_TIME;
+            columns[i++] = COL_RULE_EXEC_NOTIFICATION_ADDR;
+            columns[i++] = COL_RULE_EXEC_LAST_EXE_TIME;
+            columns[i++] = COL_RULE_EXEC_STATUS;
 
-    if ( ruleName != NULL && *ruleName != '\0' ) {
-        int i;
-        i =  genQueryInp.sqlCondInp.len;
-        /*  i2a[i]=COL_RULE_EXEC_NAME;
-        sprintf(v2,"='%s'",ruleName);  */
-        i2a[i] = COL_RULE_EXEC_ID;
-        snprintf( v2, BIG_STR, "='%s'", ruleName );
-        condVal[i] = v2;
-        genQueryInp.sqlCondInp.len++;
-    }
-
-    genQueryInp.condInput.len = 0;
-
-    genQueryInp.maxRows = 50;
-    genQueryInp.continueInx = 0;
-    status = rcGenQuery( Conn, &genQueryInp, &genQueryOut );
-    if ( status == CAT_NO_ROWS_FOUND ) {
-        if ( ruleName != NULL && *ruleName != '\0' ) {
-            printf( "User %s or rule '%s' does not exist.\n", name, ruleName );
+            long_format_len = i;
         }
-        else {
-            i1a[0] = COL_USER_COMMENT;
-            i2a[0] = COL_USER_NAME;
-            genQueryInp.selectInp.len = 1;
-            status = rcGenQuery( Conn, &genQueryInp, &genQueryOut );
-            if ( status == 0 ) {
-                if ( allFlag ) {
-                    printf( "No delayed rules pending\n" );
-                }
-                else {
-                    printf( "No delayed rules pending for user %s\n", name );
-                }
-                return 0;
-            }
-            printf( "User %s does not exist.\n", name );
-            return 0;
-        }
-    }
-
-    if ( allFlag ) {
-        printf( "Pending rule-executions\n" );
-    }
-    else {
-        printf( "Pending rule-executions for user %s\n", name );
-    }
-    printCount += printGenQueryResults( Conn, status, genQueryOut, columnNames, 0 );
-
-    while ( status == 0 && genQueryOut->continueInx > 0 ) {
-        genQueryInp.continueInx = genQueryOut->continueInx;
-        status = rcGenQuery( Conn, &genQueryInp, &genQueryOut );
-        if ( genQueryOut->rowCnt > 0 ) {
-            printf( "----\n" );
-        }
-        printCount += printGenQueryResults( Conn, status, genQueryOut,
-                                            columnNames, 0 );
-    }
-
-    return 0;
+    };
 }
 
-/*
-Via a general query, show rule information, brief form
-*/
-int
-showRuleExecBrief( char *name, int allFlag ) {
-    genQueryInp_t genQueryInp;
-    genQueryOut_t *genQueryOut;
-    int i1a[20];
-    int i1b[20] = {0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0};
-    int i2a[20];
-    char *condVal[10];
-    char v1[BIG_STR];
-    int i, status;
-    int printCount;
-
-    memset( &genQueryInp, 0, sizeof( genQueryInp_t ) );
-    printCount = 0;
-    i = 0;
-    i1a[i++] = COL_RULE_EXEC_ID;
-    i1a[i++] = COL_RULE_EXEC_NAME;
-
-    genQueryInp.selectInp.inx = i1a;
-    genQueryInp.selectInp.value = i1b;
-    genQueryInp.selectInp.len = i;
-
-    i2a[0] = COL_RULE_EXEC_USER_NAME;
-    sprintf( v1, "='%s'", name );
-    condVal[0] = v1;
-
-    genQueryInp.sqlCondInp.inx = i2a;
-    genQueryInp.sqlCondInp.value = condVal;
-    genQueryInp.sqlCondInp.len = 1;
-    if ( allFlag ) {
-        genQueryInp.sqlCondInp.len = 0;
-    }
-
-    genQueryInp.condInput.len = 0;
-
-    genQueryInp.maxRows = 50;
-    genQueryInp.continueInx = 0;
-    status = rcGenQuery( Conn, &genQueryInp, &genQueryOut );
-    if ( status == CAT_NO_ROWS_FOUND ) {
-        i1a[0] = COL_USER_COMMENT;
-        i2a[0] = COL_USER_NAME;
-        genQueryInp.selectInp.len = 1;
-        status = rcGenQuery( Conn, &genQueryInp, &genQueryOut );
-        if ( status == 0 ) {
-            if ( allFlag ) {
-                printf( "No delayed rules pending\n" );
-            }
-            else {
-                printf( "No delayed rules pending for user %s\n", name );
-            }
-            return 0;
-        }
-        if ( status == CAT_NO_ROWS_FOUND ) {
-            printf( "User %s does not exist.\n", name );
-            return 0;
-        }
-    }
-    printf( "id     name\n" );
-    printCount += printGenQueryResults( Conn, status, genQueryOut, NULL, 1 );
-
-    while ( status == 0 && genQueryOut->continueInx > 0 ) {
-        genQueryInp.continueInx = genQueryOut->continueInx;
-        status = rcGenQuery( Conn, &genQueryInp, &genQueryOut );
-        printCount += printGenQueryResults( Conn, status, genQueryOut,
-                                            NULL, 1 );
-    }
-    return 0;
-}
+auto show_RuleExec( char *userName,
+                    const char *ruleName="",
+                    int allFlag = false,
+                    bool brief = false ) -> int;
 
 int
 main( int argc, char **argv ) {
 
     signal( SIGPIPE, SIG_IGN );
 
-    int status, nArgs;
     rErrMsg_t errMsg;
 
     rodsArguments_t myRodsArgs;
@@ -267,7 +134,7 @@ main( int argc, char **argv ) {
 
     rodsLogLevel( LOG_ERROR );
 
-    status = parseCmdLineOpt( argc, argv, "alu:vVh", 0, &myRodsArgs );
+    int status = parseCmdLineOpt( argc, argv, "alu:vVh", 0, &myRodsArgs );
     if ( status ) {
         printf( "Use -h for help\n" );
         return 1;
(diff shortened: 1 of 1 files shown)


d-w-moore:
may conflict with https://github.com/irods/irods_client_icommands/commit/88fd0a9f2afd4b538fa7bfa115afa1edf0cd9beb

korydraughn:
Is this ready for review?

d-w-moore:
@korydraughn Yes. I have yet to run through the entire test suite, but comments are welcome.

src/iqstat.cpp
```diff
@@ -363,3 +241,94 @@ void usage() {
...
+                    bool brief ) -> int
+{
+    genQueryInp_t genQueryInp{};
+    genQueryOut_t *genQueryOut{};
+
+    auto num_cols_selected = (brief ? once_instance.brief_format_len
+                                    : once_instance.long_format_len);
+
+    for (int i = 0; i < num_cols_selected; i++) {
+        addInxIval(&genQueryInp.selectInp,i1a[i],0);
```
korydraughn, line 257: Let's add a space after the commas for readability.

Please apply to the other lines you've modified.

src/iqstat.cpp
```diff
@@ -363,3 +241,94 @@ void usage() {
...
+    }
+
+    if (!allFlag) {
+        snprintf( v1, BIG_STR, "='%s'", userName );
+        addInxVal(&genQueryInp.sqlCondInp,COL_RULE_EXEC_USER_NAME,v1);
+    }
+
+    std::string diagnostic {""};
```
korydraughn, line 265: The default constructor of a `std::string` results in an empty string.

This line is equivalent to: `std::string diagnostic;`

src/iqstat.cpp
```diff
@@ -363,3 +241,94 @@ void usage() {
...
+        return std::stoul(k);
+    }
+    catch(const std::invalid_argument&){
+        THROW(SYS_INVALID_INPUT_PARAM, "Delayed task ID has incorrect format.");
+    }
+    catch(const std::out_of_range&){
+        THROW(SYS_INVALID_INPUT_PARAM, "Delayed task ID is too large.");
+    }
+    catch(...){
+        THROW(SYS_INVALID_INPUT_PARAM, "Unknown error parsing task ID.");
```
korydraughn, line 332: Let's replace `task` with `rule` since that is the most common terminology used to describe a delay rule.

src/iqstat.cpp
```diff
@@ -363,3 +241,94 @@ void usage() {
...
+    while ( status == 0 && genQueryOut->continueInx > 0 ) {
+        genQueryInp.continueInx = genQueryOut->continueInx;
+        status = rcGenQuery( Conn, &genQueryInp, &genQueryOut );
+        if (!brief && genQueryOut->rowCnt > 0 ) {
+            printf( "----\n" );
+        }
+        increment_count_and_print_results();
+        clearGenQueryOut( genQueryOut );
+    }
+    clearGenQueryInp( &genQueryInp );
```
korydraughn, line 314: Can the `genQueryOut` variable result in a leak if we never enter the loop?
Can it leak after exiting the loop?

korydraughn (commented):
Looking good.

src/iqstat.cpp
```diff
@@ -17,11 +24,13 @@ char cwd[BIG_STR];
 
 int debug = 0;
 
-rcComm_t *Conn;
+rcComm_t *Conn {};
 rodsEnv myEnv;
 
 void usage();
 
+auto genQuery_id_string_to_ulong(const std::string & s) -> unsigned long;
```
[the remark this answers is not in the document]
d-w-moore: Do we have a convenient size type that's a synonym for the SQL integer key type?  Isn't it well understood to be in the range MIN_{UN,}SIGNED_LONG -> MAX_{UN,}SIGNED_LONG ?  

I can use `size_t`  or `ssize_t` if we think one of those is best.

src/iqstat.cpp
```diff
@@ -77,188 +86,53 @@ printGenQueryResults( rcComm_t *Conn, int status, genQueryOut_t *genQueryOut,
...
-        int i;
-        i =  genQueryInp.sqlCondInp.len;
-        /*  i2a[i]=COL_RULE_EXEC_NAME;
-        sprintf(v2,"='%s'",ruleName);  */
-        i2a[i] = COL_RULE_EXEC_ID;
-        snprintf( v2, BIG_STR, "='%s'", ruleName );
-        condVal[i] = v2;
-        genQueryInp.sqlCondInp.len++;
-    }
+    struct once {
```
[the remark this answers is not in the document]
d-w-moore: `formatting_info`perhaps?  Best I can do right now.
Looking now, it's only used by `show_RuleExec`, so no need for it to be global anymore. It was used by two different routines before they were merged into the one.

src/iqstat.cpp
d-w-moore: Ok. Change "Delayed task" to "Delay rule" on both lines then.  LMK if something else occurs that is better.

src/iqstat.cpp
d-w-moore: Looks like it.  Not a serious leak, but a `scoped_exit` thing might help.

d-w-moore: Well, now I  see how it can be done with at_scope_exit .
  - clearGenQueryOut is the problem. it's called at the end of the loop
  - it is not idempotent
  - i can't compensate in the lambda by setting genQueryOut to a nullptr b/c it's tested in the loop condition
 
I just don't think it's cleanly possible.  But I'm open to suggestions.

d-w-moore: One idea: possibly set / maintain a flag, like a ` gq_out_dirty` or similar name, to indicate whether the RAII should call  clearGenQueryOut.  It'd cloud the logic a tiny bit, but I suppose necessarily so, if we want to avoid the leak.

src/iqstat.cpp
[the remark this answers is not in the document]
korydraughn: I'm thinking along the lines of a type such as std::int32_t or the 64 bit variant.

Unsigned ints can wrap and hide bugs so keep that in mind as you consider signed vs unsigned.

src/iqstat.cpp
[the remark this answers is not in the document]
korydraughn: Sure. Continue to think on it.

src/iqstat.cpp
korydraughn: Perhaps the clearing of the genQueryOut variable needs to be moved just before invoking rcGenQuery.

And you add another call to clear after exiting the loop.

d-w-moore: Seems to compile now. Will check as to function in the testing phase, which will be soon. May need to write at least one more test, I'm not sure the iqstat tests i've written are stressing it enough yet.... ; )

src/iqstat.cpp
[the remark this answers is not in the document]
d-w-moore: Done!

src/iqstat.cpp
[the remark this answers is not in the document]
d-w-moore: Shored this one up, I think.

src/iqstat.cpp
```diff
@@ -357,3 +227,116 @@ void usage() {
     }
     printReleaseInfo( "iqstat" );
 }
+
+auto show_RuleExec( char *userName,
+                    const char *ruleName,
+                    int allFlag,
+                    bool brief ) -> int
+{
+    int status;
```
alanking, line 236: Please move declaration to the point of initialization

src/iqstat.cpp
```diff
@@ -357,3 +227,116 @@ void usage() {
...
+        return std::stoul(k);
+    }
+    catch(const std::invalid_argument&){
+        THROW(SYS_INVALID_INPUT_PARAM, "Delay rule ID has incorrect format.");
+    }
+    catch(const std::out_of_range&){
+        THROW(SYS_INVALID_INPUT_PARAM, "Delay rule ID is too large.");
+    }
+    catch(...){
+        THROW(SYS_INVALID_INPUT_PARAM, "Unknown error parsing delay rule ID.");
```
alanking, line 340: `SYS_UNKNOWN_ERROR`?

Consider, also, a `std::exception` case for the "known" cases.

src/iqstat.cpp
```diff
@@ -357,3 +227,116 @@ void usage() {
...
+
+    auto num_cols_selected = (brief ? output_format.brief_format_len
+                                    : output_format.long_format_len);
+
+    for (int i = 0; i < num_cols_selected; i++) {
+        addInxIval(&genQueryInp.selectInp, output_format.columns[i], 0);
+    }
+
+    char v1[BIG_STR];
+    char v2[BIG_STR];
```
alanking, line 250: Scope could be limited on these since we `strdup` the values in `addInxVal`.

src/iqstat.cpp
```diff
@@ -357,3 +227,116 @@ void usage() {
     }
     printReleaseInfo( "iqstat" );
 }
+
+auto show_RuleExec( char *userName,
+                    const char *ruleName,
+                    int allFlag,
+                    bool brief ) -> int
```
alanking, line 234: This function always returns 0. Not sure what it means, but just pointing that out as we consider the return type for this

src/iqstat.cpp
```diff
@@ -357,3 +227,116 @@ void usage() {
...
+    // RAII for both GenQuery variables.
+    auto clean_gq_in_out = [&, &i = genQueryInp, &o = genQueryOut]{ clearGenQueryInp(&i);
+                                                                    clear_gq_out();
+                                                                    free(o); o = nullptr; };
+    // Implement RAII:
+    const auto clean_gq = irods::at_scope_exit{[&] { clean_gq_in_out(); }};
+
+    status = rcGenQuery( Conn, &genQueryInp, &genQueryOut );
+
+    if ( status == CAT_NO_ROWS_FOUND ) {
```
alanking, line 278: Should we handle other `status < 0` case(s)?

alanking (commented):
Fully recognizing that this is a request to rewrite the whole iCommand, is there any way we can re-implement this using `irods::query`? :)

src/iqstat.cpp
```diff
@@ -277,54 +143,58 @@ main( int argc, char **argv ) {
...
-        status = showRuleExec( userName, argv[myRodsArgs.optind],
-                               myRodsArgs.all );
-    }
-    else {
-        if ( myRodsArgs.longOption == 0 ) {
-            status = showRuleExecBrief( userName, myRodsArgs.all );
-        }
-        else {
-            status = showRuleExec( userName, "", myRodsArgs.all );
+        if (nArgs = argc - myRodsArgs.optind; nArgs > 0) {
```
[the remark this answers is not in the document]
korydraughn: I agree.

d-w-moore:
Maybe folks can review the new help text I pushed up in the very last commit.  Meanwhile , I'll take care of the review comments  above from @alanking

src/iqstat.cpp
```diff
@@ -208,12 +208,15 @@ Print the main usage/help information.
 void usage() {
     char *msgs[] = {
         "Usage: iqstat [-luvVh] [-u user] [ruleId]",
-        "Show information about your pending iRODS rule executions",
-        "or for the entered user.",
-        " -a        display requests of all users",
-        " -l        for long format",
-        " -u user   for the specified user",
-        " ruleId for the specified rule (implies long format)",
+        "Show information about pending iRODS rule executions (i.e. delayed rules).",
```
korydraughn, line 211: Let's separate the USAGE line from the description. You can do that using the same idea as on L214.

src/iqstat.cpp
```diff
@@ -208,12 +208,15 @@ Print the main usage/help information.
...
+        "Listings are produced via a general query of the R_RULE_EXEC table.  By default,",
```
korydraughn, line 212: Let's remove the first statement here. The reader doesn't need to know how the information is gathered.

Keep the statement about the default behavior though.

src/iqstat.cpp
```diff
@@ -357,3 +227,116 @@ void usage() {
...
+auto show_RuleExec( char *userName,
+                    const char *ruleName,
+                    int allFlag,
+                    bool brief ) -> int
+{
+    int status;
+    genQueryInp_t genQueryInp{};
+    genQueryOut_t *genQueryOut{};
+
+    output_format_parameters output_format;
```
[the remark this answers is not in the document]
d-w-moore: I guess they could - it didn't occur to me for such cases.  Is there a general rationale for `constexpr` in the case of stack variables?

src/iqstat.cpp
```diff
@@ -208,12 +208,15 @@ Print the main usage/help information.
...
-        " ruleId for the specified rule (implies long format)",
+        "Show information about pending iRODS rule executions (i.e. delayed rules).",
+        "Listings are produced via a general query of the R_RULE_EXEC table.  By default,",
+        "only the invoking user's rules are shown.",
+        " ",
+        "Options are:",
+        " -a        display requests of all users.",
+        " -l        force use of the long format.",
+        " -u user   display requests for the specified user (may not be used together with -a).",
+        " ruleId    limits the query to the specified rule ID (and implies long format).",
```
korydraughn, line 219: Let's capitalize the first letter of each description.

src/iqstat.cpp
```diff
@@ -208,12 +208,15 @@ Print the main usage/help information.
 void usage() {
     char *msgs[] = {
         "Usage: iqstat [-luvVh] [-u user] [ruleId]",
```
korydraughn, line 210: Consider making this better. Take a look at `istream` and `itouch`'s help text.

This is not required. Feel free to ignore this comment.

src/iqstat.cpp
```diff
@@ -277,54 +143,58 @@ main( int argc, char **argv ) {
...
-                               myRodsArgs.all );
-    }
-    else {
-        if ( myRodsArgs.longOption == 0 ) {
-            status = showRuleExecBrief( userName, myRodsArgs.all );
-        }
-        else {
-            status = showRuleExec( userName, "", myRodsArgs.all );
+        if (nArgs = argc - myRodsArgs.optind; nArgs > 0) {
+            static_cast<void>(genQuery_id_string_to_id( argv[myRodsArgs.optind]));
```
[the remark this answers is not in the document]
d-w-moore: The compiler didn't like the unused variable without the void cast.  The result isn't important, only that an error is not thrown by an ID parameter that is incorrectly formatted.

src/iqstat.cpp
d-w-moore: Done

src/iqstat.cpp
d-w-moore: Done

src/iqstat.cpp
d-w-moore: Not sure the best approach.  Should the error stack be printed in that case.  Should the status < 0 be converted to an irods excption and then thrown for the main( ) function to catch ?

src/iqstat.cpp
d-w-moore: Yep. There were two functions originally, showRuleExec and showRuleExecBrief.  They both always returned 0.  This function is the synthesis of those two.

src/iqstat.cpp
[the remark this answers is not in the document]
korydraughn: So this is checking whether `argv[myRodsArgs.optind]` can be converted to an integer?

Is the cast still needed even if we're not capturing the result of the call?

src/iqstat.cpp
d-w-moore: I'll give it a go.  oh yes, also, I was conflicted w.r.t  the verbose options.  Due to the way they're processed, `-v` is accepted but does nothing, and `-V` has effects not evident within the client's source code, effects which I feel unprepared to describe, other than "makes the client very verbose."

src/iqstat.cpp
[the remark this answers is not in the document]
d-w-moore: Trying it again , the compiler appears to no longer flag it as an error. I'll remove the cast

src/iqstat.cpp
d-w-moore: Done

src/iqstat.cpp
d-w-moore: done!

src/iqstat.cpp
d-w-moore: made the arguments stand out a little better via all capitals and aligned everything better, like the aforementioned sample icommand help texts.  Here's what is produced now:
```
$ ./iqstat -h
Usage:     iqstat [-luvVh] [-u USER] [RULE_ID]
 
Show information about pending iRODS rule executions (i.e. delayed rules).
By default, only the invoking user's rules are shown.
 
Options:
 -a        Display requests of all users.
 -l        Force use of the long format.
 -u USER   Display requests for the specified user (may not be used together with -a).
 RULE_ID   Limits the query to the specified integer rule ID (and implies long format).
 
See also iqdel and iqmod

iRODS Version 4.2.11                iqstat

```

src/iqstat.cpp
d-w-moore: done!

src/iqstat.cpp
d-w-moore: ... or is the std::exception case more likely (or guaranteed even) to contain an error string.  That would make sense I guess.  Wll try it that way then.

d-w-moore: Ok, done ... i think.

src/iqstat.cpp
korydraughn: I like it. Mark this comment as resolved when you're ready.

d-w-moore: coolness

src/iqstat.cpp
alanking: I was just commenting on the fact that there does not seem to be any error handling here and it may be worth just printing a message and getting out if `status < 0` (and is `!= CAT_NO_ROWS_FOUND`, of course)

src/iqstat.cpp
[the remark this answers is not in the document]
alanking: Mmmm, not really. I just like it when things are nailed down :)

src/iqstat.cpp
alanking: Okay. Looking at this again, this is sort of related to my comment further down about error handling. Feel free to resolve this in favor of the other.

src/iqstat.cpp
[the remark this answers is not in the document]
d-w-moore: Yep, sounds good.

src/iqstat.cpp
d-w-moore: Seems good.

src/iqstat.cpp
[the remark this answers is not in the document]
d-w-moore: Done

src/iqstat.cpp
d-w-moore: ok. closing in favor of other solution, handled below.

src/iqstat.cpp
d-w-moore: Ok, think I  took care of this one. Please review change and lmk.

src/iqstat.cpp
```diff
@@ -1,30 +1,43 @@
 #include "irods_client_api_table.hpp"
 #include "irods_pack_table.hpp"
+#include "irods_query.hpp"
+#include "rcMisc.h"
 #include "rods.h"
 #include "rodsClient.h"
+#include "user_administration.hpp"
+#include "rodsLog.h"
```
alanking, line 8: Alphabetize

src/iqstat.cpp
```diff
@@ -357,3 +234,137 @@ void usage() {
...
+    else if (status < 0) {
+        char* errno_message{};
+        const char* main_message = rodsErrorName(status, &errno_message);
+        std::cerr << "Error in rcGenQuery: " << main_message;
+        if (errno_message && *errno_message) {
+            std::cerr << " - "s + errno_message;
+        }
+        std::cerr << "\n";
+        std::free(errno_message);
+        return 0;
```
alanking, line 317: Should this be something other than 0 to indicate the failure?

src/iqstat.cpp
```diff
@@ -357,3 +234,137 @@ void usage() {
...
+            return 0;
+        }
+    }
+    else if (status < 0) {
+        char* errno_message{};
+        const char* main_message = rodsErrorName(status, &errno_message);
+        std::cerr << "Error in rcGenQuery: " << main_message;
+        if (errno_message && *errno_message) {
+            std::cerr << " - "s + errno_message;
+        }
```
alanking, line 314: Will the error message end up looking like this if `errno_message` has contents?
`"Error in rcGenQuery:  - <message goes here>"`
Looks kind of funny to me. Maybe the " - " or the colon following the message on 311 can be omitted?

alanking (commented):
Very, very close now...

src/iqstat.cpp
```diff
@@ -276,58 +144,64 @@ main( int argc, char **argv ) {
...
-        status = showRuleExec( userName, argv[myRodsArgs.optind],
-                               myRodsArgs.all );
-    }
-    else {
-        if ( myRodsArgs.longOption == 0 ) {
-            status = showRuleExecBrief( userName, myRodsArgs.all );
-        }
-        else {
-            status = showRuleExec( userName, "", myRodsArgs.all );
+        int nArgs { argc - myRodsArgs.optind };
```
alanking, line 187: Unless I'm missing something, I think your `try` block can start right above this line here. Nothing above this throws as far as I know... Would reduce your diff a little bit, too. Up to you.

src/iqstat.cpp
d-w-moore: Oh, you are right.  Do we have any set correspondence between return values and their meaning? Just curious.  I've seen lots of return 1; 2; or 3, but never any rationale.

src/iqstat.cpp
d-w-moore: I'm sure Error in rcGenQuery will always have an all caps error code immediately after, which is contained in the main_message variable.  

Do we have a convention for how the errno sub-message is printed?  `ERRCODE ( errno-msg )` or maybe `ERRCODE: errno-msg` ?

src/iqstat.cpp
d-w-moore: Yes, doing that....

src/iqstat.cpp
alanking: Sorry, I meant to say "if `errno_message` has *no* contents". In any case, if that happens, then something has gone really sideways, based on what you're saying. This is fine as-is, then.

src/iqstat.cpp
d-w-moore: yep

src/iqstat.cpp
d-w-moore: I think this is working now more in the manner of other icommands' server status reporting. Tested manually using gdb.

src/iqstat.cpp
d-w-moore: Returning 1 similar to handling of other iRODS errors.

korydraughn: That will depend on the icommand, but in general, no.

If the meaning of the error code cannot be derived, use 1. It's more important that the user receive a non-zero error code than not at all.

src/iqstat.cpp
```diff
@@ -276,58 +144,64 @@ main( int argc, char **argv ) {
         usage();
         return 0;
     }
-
     status = getRodsEnv( &myEnv );
     if ( status < 0 ) {
-        rodsLog( LOG_ERROR, "main: getRodsEnv error. status = %d",
-                 status );
-        return 1;
+	// TODO might have to change this for 4.3.0
```
alanking, line 149: I think this comment can be removed as this probably just won't compile in 4.3.0 and doesn't add much value in 4-2-stable (since it's not 4.3 :))

trel: yes, no mention of 4.3 in 4-2-stable please.

d-w-moore: Deleting. I believe this actually might compile in 4.3 assuming the include were changed to <irods/rodsLog.h>.  But, placing the TODO in my notes instead to output directly to stderr

d-w-moore: squash done

src/iqstat.cpp
```diff
@@ -1,30 +1,43 @@
...
 
 int debug = 0;
 
-rcComm_t *Conn;
+rcComm_t *Conn{};
 rodsEnv myEnv;
 
 void usage();
 
+auto throw_if_id_cannot_be_converted_to_int(const std::string& key) -> void;
```
korydraughn, line 31: Needs the std::string header.

d-w-moore: check.

src/iqstat.cpp
```diff
@@ -1,30 +1,43 @@
...
+#include <boost/format.hpp>
+
+#include <cstdint>
+#include <cstdio>
+#include <stdexcept>
 
 #define MAX_SQL 300
 #define BIG_STR 200
 
+using namespace std::string_literals;
```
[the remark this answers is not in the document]
d-w-moore: Removed, but including <string> anyway as noted later on down.

src/iqstat.cpp
```diff
@@ -276,58 +144,62 @@ main( int argc, char **argv ) {
...
+	return 2;
     }
 
     if ( myRodsArgs.user ) {
-        snprintf( userName, sizeof( userName ), "%s", myRodsArgs.userString );
+	snprintf( userName, sizeof( userName ), "%s", myRodsArgs.userString );
     }
     else {
-        snprintf( userName, sizeof( userName ), "%s", myEnv.rodsUserName );
+	snprintf( userName, sizeof( userName ), "%s", myEnv.rodsUserName );
```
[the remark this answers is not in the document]
d-w-moore: done

alanking (commented):
Looks good to me

korydraughn:
Code looks good. Please squash to taste and run through the testing environment.

d-w-moore:
This has been put through the test suite on ubuntu18 on `4-2-stable` and `main`. All tests passed for both branches.

korydraughn (approved):
Excellent! Please add the pound.

d-w-moore:
Ok - pound added.

PR #353, opened 2023-02-02
[irods/irods#6887] send scrambled password only if param not empty (4-2-stable)

d-w-moore:
Corresponds to changes in https://github.com/irods/irods/pull/6889 . `igroupadmin` was scrambling an empty password string when the  password parameter was not provided; this fixes that oversight.

### DIFF
--- a/src/igroupadmin.cpp
+++ b/src/igroupadmin.cpp
@@ -291,7 +291,8 @@ doCommand( char *cmdToken[] ) {
     }
 
     if ( strcmp( cmdToken[0], "mkuser" ) == 0 ) {
-        userAdmin( "mkuser", cmdToken[1], setScrambledPw( cmdToken[2] ),
+        char empty_string[] = "";
+        userAdmin( "mkuser", cmdToken[1], *cmdToken[2] ? setScrambledPw(cmdToken[2]) : empty_string,
                    cmdToken[3], "", "", "", "" );
         return 0;
     }
@@ -333,8 +334,8 @@ main( int argc, char **argv ) {
 
     int argOffset;
 
-    int maxCmdTokens = 20;
-    char *cmdToken[20];
+    constexpr int maxCmdTokens = 20;
+    char *cmdToken[maxCmdTokens];
     int keepGoing;
     int firstTime;
 


src/igroupadmin.cpp
```diff
@@ -290,8 +290,9 @@ doCommand( char *cmdToken[] ) {
         return 0;
     }
 
+    char empty_string[] = "";
```
korydraughn, line 293: Looks like this can be moved inside of the if-stmt.

src/igroupadmin.cpp
```diff
@@ -333,8 +334,8 @@ main( int argc, char **argv ) {
 
     int argOffset;
 
-    int maxCmdTokens = 20;
-    char *cmdToken[20];
+    const int maxCmdTokens = 20;
```
korydraughn, line 337: `const` -> `constexpr`

src/igroupadmin.cpp
```diff
@@ -333,8 +334,8 @@ main( int argc, char **argv ) {
...
+    char *cmdToken[maxCmdTokens];
```
korydraughn, line 338: Do we need to clear the array?

korydraughn (commented):
Looking good.

src/igroupadmin.cpp
d-w-moore: That happens a few lines down, each element gets an empty string by default. https://github.com/d-w-moore/irods_client_icommands/blob/1815fab27082a02d5cb36c1123953ffc11bf8c11/src/igroupadmin.cpp#L364

korydraughn: Good enough for me.

src/igroupadmin.cpp
d-w-moore: done

src/igroupadmin.cpp
d-w-moore: done

src/igroupadmin.cpp
```diff
@@ -291,7 +291,8 @@ doCommand( char *cmdToken[] ) {
     }
 
     if ( strcmp( cmdToken[0], "mkuser" ) == 0 ) {
-        userAdmin( "mkuser", cmdToken[1], setScrambledPw( cmdToken[2] ),
+        char empty_string[] = "";
+        userAdmin( "mkuser", cmdToken[1], *cmdToken[2] ? setScrambledPw(cmdToken[2]) : empty_string,
```
korydraughn, line 295: Can `cmdToken[2]` be null?
Is `cmdToken[2]` guaranteed to be null-terminated?

d-w-moore: For cmdToken in the client to be both non nullptr in value and null terminated depends on both: 
   - its assigned `argv` value to be null terminated, which they always are, assuming we don't scan past `argc` # of values on the cmdline (which we don't); backed up, of course, by the initialization of unassigned values to "" in the loop (in `main`, and in `getInput` for interactive mode) calls preceding the assignment.  
   - `getInput( )` to always guarantee said conditions, which a fairly thorough look-through seems to confirm.

However, I wouldn't argue against the possibility of improvement

korydraughn: I see. The pointers of `cmdToken` always point to strings in `argv` therefore they are guaranteed to be null-terminated. That means checking the first byte for zero is acceptable as a check for empty string.

Good enough for me. Please mark as resolved when ready.

alanking (commented):
I think this can be squashed down, omitting #s

korydraughn:
Yes, please squash without the #s.

korydraughn (approved):
Please add the pound.

alanking (approved):

PR #352, opened 2023-01-30
[irods/irods_5411] Allow longer `HOME` directory in iinit

MartinFlores751:
Cherry-picked from #336.

PR #354, opened 2023-02-09
[irods/irods_5411] Allow longer `HOME` directory in iinit (4-2-stable)

MartinFlores751:
Enable `iinit` to handle longer `$HOME` directory.

Cherry-pick of #336 or #352.

This did change a bit, since the filesystem library is not available for 4-2-stable.

### DIFF
--- a/src/iinit.cpp
+++ b/src/iinit.cpp
@@ -13,12 +13,14 @@
 #include "irods_environment_properties.hpp"
 #include "irods_kvp_string_parser.hpp"
 
+#include "boost/filesystem.hpp"
 #include "boost/lexical_cast.hpp"
 
 #include "json.hpp"
+#include <fmt/format.h>
 
-#include <iostream>
 #include <fstream>
+#include <iostream>
 
 void usage( char *prog );
 void usageTTL();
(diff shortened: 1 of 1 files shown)


korydraughn:
We also have Boost.FileSystem in case you weren't aware.

Up to you whether you want to use it or not.

MartinFlores751:
Did not know that, I'll try it out and see how it works.

Think it might be a bit more readable than having something like `S_IRWXU`.

korydraughn:
Go for it.

We package various things for iRODS. Boost is one of them. Assuming you have the package repository configured, you can see which packages are available by running `apt-cache search 'irods-externals*'`.

src/iinit.cpp
```diff
@@ -14,11 +14,13 @@
 #include "irods_kvp_string_parser.hpp"
 
 #include "boost/lexical_cast.hpp"
+#include "boost/filesystem.hpp"
```
alanking, line 17: Alphabetize

alanking (commented):
Looks good to me. Please squash when ready (and you can address my comment, or not :p)

alanking:
Please squash here as well

alanking (approved):
Yes, please add #

korydraughn (approved):

PR #358, opened 2023-02-17
[irods/irods#6887] send scrambled password only if param not empty (4-3-stable)

d-w-moore:


d-w-moore:
delete, this is another one for "wrong branch"

PR #360, opened 2023-02-20
[irods/irods#6896] irm: -n option should not be listed (main)

alanking:
I've lumped this into the effort to address irods/irods#6896, but I'm open to making a dedicated issue for it.


trel:
oh!  we just missed that usage text.   yes, this is fine.

korydraughn (approved):
Ship it!

PR #361, opened 2023-02-20
[irods/irods#6896] irm: -n option should not be listed (4-3-stable)

alanking:
In service of irods/irods#6896

Cherry-picked from #360


korydraughn (approved):
Pound and merge when ready.

PR #362, opened 2023-02-21
[irods/irods#6888] extra help text documentation for igroupadmin (4-2-stable)

d-w-moore:


src/igroupadmin.cpp
```diff
@@ -466,14 +466,25 @@ void usageMain() {
         "certain administrative functions.  It can also be used by regular",
         "users to list groups ('lg') and by the admin for all operations.",
         " ",
+        "After making a new group they will themselves manage, a groupadmin's
+        "first action should be to add their own username to the group.",
```
alanking, line 470: Maybe we should leave an example of how to do this, e.g.:
```bash
$ igroupadmin atg my_new_empty_group groupadmin_username
```

src/igroupadmin.cpp
```diff
@@ -466,14 +466,25 @@ void usageMain() {
...
+        "They will from that point forward be able to add users to, and remove",
+        "them from, the group.  Those operations permit, as well, the transfer",
+        "of responsibility for the group to another groupadmin.",
+        " ",
+        "Currently, a groupadmin may not remove groups or users from the zone."
```
alanking, line 475: I'm definitely guilty of adding words like this to my stuff as well, but I think "Currently, " isn't adding much here. Consider removing it.

Missing a comma at the end of this line.

src/igroupadmin.cpp
```diff
@@ -466,14 +466,25 @@ void usageMain() {
...
+        " ",
         "A blank execute line invokes the interactive mode, where it",
         "prompts and executes commands until 'quit' or 'q' is entered.",
         "Single or double quotes can be used to enter items with blanks.",
         "Commands are:",
         " lg [name] (list group info (user member list))",
         " mkuser Name Password (make a user and set the initial password)",
         " atg groupName userName[#Zone] (add to group - add a user to a group)",
+        "   * NOTE: atg is allowed for a groupadmin only if they belong to the group given,",
+        "     and it is currently empty (no members)."
```
alanking, line 485: Is this true? Can't the groupadmin add users to the group if they are a member of the group? Or is it only when they are the *first* member in a group?

Please also consider moving this text to the `atgMsgs` which gives more details about the `atg` subcommand. Users can view this text with `igroupadmin h atg` and the text will appear on this page in the next version (I've linked the 4.3.0 page for demonstration purposes only): https://docs.irods.org/4.3.0/icommands/groupadmin/

This has been the precedent for all the iCommands which use subcommands (`iadmin`, `igroupadmin`, `imeta`, `iticket`). If we want to keep this text front and center, that's fine, just wanted to make mention of this.

Missing a comma at the end of this line.

src/igroupadmin.cpp
```diff
@@ -466,14 +466,25 @@ void usageMain() {
...
         " rfg groupName userName[#Zone] (remove from group - remove a user from a group)",
+        "   * NOTE: rfg is allowed for a groupadmin only if they belong to the group given.",
```
alanking, line 487: Same comment here as for `atg`: Please consider moving/adding this message to the `rfgMsgs`.

src/igroupadmin.cpp
alanking: Or, get really hand-holdy:
```bash
# as groupadmin "labmanager1"
$ igroupadmin mkgroup lab1
$ igroupadmin atg lab1 labmanager1
```

src/igroupadmin.cpp
alanking: Sorry, meant to say:

Can't the groupadmin add users to the group if they are a member of the group and it's not empty as well?
```bash
# as labmanager1
$ igroupadmin mkgroup lab1
$ igroupadmin atg lab1 labmanager1
$ igroupadmin atg lab1 alice
$ igroupadmin atg lab1 bobby
```
That should work (I thought) and `atg` succeeds when the group is not empty. Maybe I'm just reading the sentence wrong. :)

d-w-moore: > Is this true? Can't the groupadmin add users to the group if they are a member of the group? Or is it only when they are the _first_ member in a group?
> 
I think youre right;  will revisit

d-w-moore: I don't know about moving the NOTE annotated stuff to the subcommand help.  To me, that makes it more obscure. It's the reason i wasn't familiar with the special kinks of igroupadmin in the first place. Someone has to know there's stuff to dig for, in the output of "igroupadmin h atg" , ie. more than what one would find in "man igroupadmin",

src/igroupadmin.cpp
d-w-moore: (see atgMsgs comment above.)

src/igroupadmin.cpp
alanking: Understood. I defer to others on this matter, then.

d-w-moore: Me too tbh.  Could do with input from @trel and @korydraughn .  Do we take a dual approach, then ? Teach the philosophy in the main text and give examples within each subcommand text ?

trel: I'd be happy to get rid of the subcommand help text - seems like we can fit whatever we need at the top level... ?

korydraughn: I'm not a fan of how a person views the help text for a subcommand, but I feel we should move the note about constraints for `atg` and `rfg` into the subcommand help text.

We can include a line in the top level help text that instructs the user to view the subcommand help text of those operations.

For example:
```
Usage: igroupadmin [-vV] [command]
Usage: igroupadmin h [command]

<snip>

NOTE: atg and rfg have additional usage requirements. See the help text
of these subcommands for details (e.g. igroupadmin h atg).

Commands are:
 <snip>
 atg groupName ...
 rfg groupName ...
```

Moving all of the subcommand help text into the top level would work, but it feels incorrect to print subcommand help text at that level.

Willing to be persuaded otherwise.

trel: That seems fine with me.

korydraughn:
Looking good.

I noticed that the lines which listed the usage requirements for `atg` and `rfg` were not moved into the help text of the subcommands. I think that list should be included in the help text of the subcommands.

I can't remember what you had exactly, but I know it was something similar to the following:
```
igroupadmin h atg
<snip>
The following requirements must be satisfied in order to use this command:
  - The invoking user must be a member of the group, or
  - The invoking user must be the only member of the group, or
  - The group must be empty
<snip>
```

Thoughts?

d-w-moore:
> I noticed that the lines which listed the usage requirements for atg and rfg were not moved into the help text of the > subcommands. 
> 
> Thoughts?

I'll look back and see what else I can integrate.  When i looked at what already exists in the subcommand help text, it seemed that the bits I'd be moving in would just be repetition....

d-w-moore:
Originally I'd had the impression that a paragraph or so giving overall philosophy behind the groupadmin functions and how they work would be called for.  Then it seemed the consensus that stuff needed to be moved into the individual subcommands, but they already seem to state what users would need to know, just  ... a bit more piecemeal.  

I think perhaps we leave it as is now, unless we think some critical info is being omitted? After having gone through it, I don't think so. It does help that, with this latest commit, we are printing out the full help texts rather than just partially.

korydraughn:
That's fine. We can always improve documentation if people find things to be lacking.

korydraughn:
I looked over the help text again and I do see an issue. The help text for `atg` does not tell the user they cannot add themselves to a non-empty group. And all of the text combined does not improve the situation.

The `atg` help text needs to state its requirements clearly.

d-w-moore:
Yes, we at least have a hint in the main help now, pointing people to the subcommand help where needed.

d-w-moore:
> I looked over the help text again and I do see an issue. The help text for `atg` does not tell the user they cannot add themselves to a non-empty group. And all of the text combined does not improve the situation.
> 
> The `atg` help text needs to state its requirements clearly.

ok will take a look

d-w-moore:
That's implied by the text : "users of type 'groupadmin' can atg and rfg for groups they are members of."
So we should state it more explicitly and earlier?

d-w-moore:
Like I say, it's more piecemeal.  If you read the text for mkgroup, it makes it clear you have to add yourself first.

d-w-moore:
Which is why I thought more text was required at the top level.

korydraughn:
Ah, I misread the help text again. I agree that `mkgroup` implies that, but given that we have a note in the top level text that states usage requirements are explained in the help text of `atg` and `rfg`, we have to include the requirements in the help text of each subcommand.

d-w-moore:
Ok - hope it's a bit better now. I've also removed help cross-references to rmgroup, which doesn't exist for `igroupadmin`.

src/igroupadmin.cpp
```diff
@@ -509,26 +513,48 @@ usage( char *subOpt ) {
     char *atgMsgs[] = {
         " atg groupName userName[#userZone] (add to group - add a user to a group)",
         "For remote-zone users, include the userZone.",
-        "Also see mkgroup, rfg and rmgroup.",
+        " ",
+        "NOTE: A groupadmin may add themselves to an empty group (such as one just created",
+        "but not to one already containing other users.",
```
korydraughn, line 518: Missing a closing paren?

korydraughn (commented):
Much better.

Once the final comments are resolved, please squash without the pounds and we'll review one last time.

src/igroupadmin.cpp
```diff
@@ -509,26 +513,48 @@ usage( char *subOpt ) {
...
         "for groups they are members of.  They can see group membership via iuserinfo.",
         ""
     };
 
     char *mkgroupMsgs[] = {
         " mkgroup groupName[#Zone] (make a new group)",
-        "Make a new group.  You will need to add yourself to the new group to then",
-        "be able to add and remove others (when the group is empty groupadmins are",
-        "allowed to add themselves.)",
+        "Make a new group.  You will need to add yourself to the new group *first* in",
```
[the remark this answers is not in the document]
d-w-moore: done

src/igroupadmin.cpp
d-w-moore: done

src/igroupadmin.cpp
d-w-moore: amended

src/igroupadmin.cpp
```diff
@@ -509,26 +513,48 @@ usage( char *subOpt ) {
...
     char *rfgMsgs[] = {
         " rfg groupName userName[#userZone] (remove from group - remove a user from a group)",
         "For remote-zone users, include the userZone.",
-        "Also see mkgroup, afg and rmgroup.",
+        " ",
+        "NOTE: A groupadmin may only add or remove members of a group to which they already",
+        "belong.  Also, the rfg command can be combined with atg to transfer responsibility for",
+        "an existing group to another groupadmin. (See atg help for an example.)",
+        " ",
+        "Also see mkgroup and afg.",
```
korydraughn, line 539: afg -> atg

korydraughn (approved):
Excellent.

There is a typo in the `rfg` help text. Please fix that.

Once the typo has been resolved, mark the rest of the comments as resolved and add the pound.

src/igroupadmin.cpp
d-w-moore: Oh, thanks for noticing that....

alanking (approved):
Looks good. # it

korydraughn (approved):

alanking (approved):

PR #363, opened 2023-02-21
[irods/irods#6879] iput: Disconnect in signal handler (main)

alanking:
Addresses irods/irods#6879

I am not planning to add a test for this. See https://github.com/irods/irods/issues/2634. Please let me know if I should try to add an automated test for this.

I tried to leave extensive comments in the commit message and in the code to explain why results may be inconsistent. In the end, the data object will have a stale replica as a result of interrupting the process, and that's all we care about (I think?).

Not sure if we want to have a similar facility for `irsync` or other iCommands which put data into iRODS directly from the client.

### DIFF
--- a/src/iput.cpp
+++ b/src/iput.cpp
@@ -1,27 +1,39 @@
-/*** Copyright (c), The Regents of the University of California            ***
- *** For more information please refer to files in the COPYRIGHT directory ***/
-/*
- * iput - The irods put utility
-*/
-
-#include <irods/rodsClient.h>
-#include <irods/parseCommandLine.h>
-#include <irods/rodsPath.h>
-#include <irods/putUtil.h>
-#include <irods/rcGlobalExtern.h>
 #include <irods/irods_client_api_table.hpp>
 #include <irods/irods_pack_table.hpp>
 #include <irods/irods_parse_command_line_options.hpp>
+#include <irods/parseCommandLine.h>
+#include <irods/putUtil.h>
+#include <irods/rcGlobalExtern.h>
+#include <irods/rodsClient.h>
+#include <irods/rodsPath.h>
 
 void usage( FILE* );
+
+namespace
+{
+    rcComm_t* conn;
+
+    auto disconnect_and_exit(int _signal) -> void
+    {
+        // If the process is interrupted, make sure to disconnect from the server so that the service agent properly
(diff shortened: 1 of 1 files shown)


src/iput.cpp
```diff
@@ -1,27 +1,39 @@
...
 
 void usage( FILE* );
 
+namespace
+{
+    rcComm_t* conn;
+
+    auto disconnect_and_exit(int _signal) -> void
+    {
+        // If the process is interrupted, make sure to disconnect from the server so that the service agent properly
```
trel, line 18: "service agent" ?   perhaps "servicing agent"?

src/iput.cpp
```diff
@@ -1,27 +1,39 @@
...
+        // finalizes any dangling transfers.
+        rcDisconnect(conn);
+
+        // putUtil handles the parallel transfer threads, so the connections cannot be closed from iput. Therefore, the
+        // server will still show SYS_HEADER_READ_LEN_ERR once this process exits. There may be some inconsistency in
+        // results at this point because the writing threads may finish up in the time between disconnecting and this
+        // process exiting.
+        exit(1);
```
trel, line 26: would giving a 1s sleep help with consistency?    am I a bad person for suggesting such a thing?

trel (commented):
one comment, one questionable suggestion.

src/iput.cpp
```diff
@@ -1,27 +1,39 @@
...
+    } // disconnect_and_exit
```
korydraughn, line 27: Keep in mind we are only allowed to use async-signal-safe functions in signal handlers.

See https://man7.org/linux/man-pages/man7/signal-safety.7.html.

Thinking about this more, I'm not sure we can resolve this without touching many things.

alanking: Excellent point.

I think because of that it might be safe to say that we can't address this issue, at least not in 4.2.12...

Thoughts?

korydraughn: Yes. Let's leave 4.2 as is and investigate handling this in 4.3.

alanking: Okay. In that case, I will close this PR and punt the issue. Thanks

alanking:
Closing in favor of some other to-be-determined solution

PR #373, opened 2023-03-12
[irods/irods#6948] Improved iquest help text regarding case-insensitive search (4-2-stable)

korydraughn:


src/iquest.cpp
```diff
@@ -51,6 +51,11 @@ usage() {
         "in the 'WHERE' condition will be made upper case so one can do case-insensitive",
         "tests (using upper-case literals).",
         " ",
+        "To do a case-insensitive search, pass the 'uppercase' (or 'upper') option and",
+        "make ALL conditional input values uppercase.  For example:",
+        " ",
+        "  iquest uppercase \"SELECT DATA_ID WHERE COLL_NAME LIKE '%/TEST' AND DATA_NAME = 'FOO.BIN'\"",
```
alanking, line 57: Not sure if it's worth noting a ~bug~ limitation, but if anything in the `WHERE` clause includes numbers (e.g. `DATA_SIZE`), GenQuery will return a `CAT_SQL_ERR` when used with `upper`/`uppercase`.

alanking (approved):
Approving as-is with a suggestion which may be ignored

src/iquest.cpp
korydraughn: Yes, will definitely include that.

alanking: Sorry, I meant to say if any of the conditions are supposed to be interpreted as numbers. If a `DATA_NAME`, for instance, has a number, that's okay. I think you got it. :p

korydraughn: Done. Please see new commit for new text.

trel: words seem good.

is it too much to make it NOT break in that way?   or that's for later...

alanking: I discovered this issue when investigating fixing `upper`/`uppercase` for `DATA_RESC_HIER` in https://github.com/irods/irods/issues/6912. I guess we never had a formal conversation/decision-making process about it.

My vote is that this will be better in GenQuery2 and we leave it alone for GenQuery1.

korydraughn: I agree that we leave GenQuery1 as-is. GenQuery2 is right around the corner and things like this give users a reason to reach for it and provide feedback.

trel: yeps - all good.   document and move on down the line.

src/iquest.cpp
```diff
@@ -51,6 +51,15 @@ usage() {
...
+        " ",
+        "Case-insensitive search will fail if the WHERE-clause contains columns which have",
+        "an integer-based data type.  That means, columns such as DATA_SIZE will cause the",
+        "query to fail with a CAT_SQL_ERR.",
```
korydraughn, line 61: It's probably safer/better to say that all columns in the where-clause need to have a string-based type. As it is, some users may think that ONLY integers break the query.

So how about ...
```
Case-insensitive search requires that all columns in the WHERE-clause to have a
string-based data type (e.g. varchar).  That means, columns such as DATA_SIZE
will cause the query to fail with a CAT_SQL_ERR.
```

trel: that does seem more inclusive / correct.

korydraughn: Done.

alanking (approved):

korydraughn:
Added #

korydraughn:
Hold on merging this.

src/iquest.cpp
```diff
@@ -47,9 +47,14 @@ usage() {
         "If 'no-distinct' appears before the query_string, the normal distinct option on",
         "the SQL will be bypassed (this is useful in rare cases).",
         " ",
-        "If 'uppercase' (or 'upper') appears before the query_string, the database value",
-        "in the 'WHERE' condition will be made upper case so one can do case-insensitive",
-        "tests (using upper-case literals).",
```
korydraughn, line 52: Removed the old text about case-insensitive search.

Thoughts?

korydraughn: Great. I'll add the pound.

korydraughn:
Ready for merging.

PR #379, opened 2023-03-15
[#6136] iexit: avoid removing auth file for service account (unless forced) (main)

SwooshyCueb:
Prerequisite: irods/irods#6969
Addresses irods/irods#6136 for main

Will backport to stable branches after reviews done.

### DIFF
--- a/src/iexit.cpp
+++ b/src/iexit.cpp
@@ -1,8 +1,10 @@
-/*** Copyright (c), The Regents of the University of California            ***
- *** For more information please refer to files in the COPYRIGHT directory ***/
-#include <irods/rods.h>
+#include <irods/irods_exception.hpp>
+#include <irods/irods_service_account.hpp>
 #include <irods/parseCommandLine.h>
 #include <irods/rcMisc.h>
+#include <irods/rods.h>
+
+#include <cstdio>
 
 void usage( char *prog );
 
@@ -13,7 +15,7 @@ main( int argc, char **argv ) {
 
     rodsArguments_t myRodsArgs;
 
-    int status = parseCmdLineOpt( argc, argv, "Vvh", 0, &myRodsArgs );
+    int status = parseCmdLineOpt(argc, argv, "fVvh", 0, &myRodsArgs);
     if ( status ) {
         printf( "Use -h for help\n" );
         exit( 1 );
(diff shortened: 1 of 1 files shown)


src/iexit.cpp
```diff
@@ -26,12 +35,39 @@ main( int argc, char **argv ) {
...
     char* envFile = getRodsEnvFileName();
-    status = unlink( envFile );
-    if ( myRodsArgs.verbose == True ) {
-        printf( "Deleting (if it exists) session envFile: %s\n", envFile );
-        printf( "unlink status [%d]\n", status );
-        // prompt before removing
-        obfRmPw( 0 );
+    const bool isServiceAccount = irods::is_service_account();
+
+    status = unlink(envFile);
```
trel, line 40: Do we want to wrap the `unlink` itself with `!isServiceAccount`?

That would prevent the environment file from being deleted too... 

I'm feeling a markdown table itching to get out...

SwooshyCueb: I don't quite understand the logic going on behind the scenes, but when `iexit` is run as the service account, it doesn't try to remove `irods_environment.json`, but `irods_environment.json.[SOME NUMBER]`

trel: oooh, right.   because there could be multiple connections/sessions for the same user.

carry on.

src/iexit.cpp
```diff
@@ -26,12 +35,39 @@ main( int argc, char **argv ) {
     rodsEnv myEnv;
     getRodsEnv( &myEnv );
     char* envFile = getRodsEnvFileName();
-    status = unlink( envFile );
-    if ( myRodsArgs.verbose == True ) {
-        printf( "Deleting (if it exists) session envFile: %s\n", envFile );
-        printf( "unlink status [%d]\n", status );
-        // prompt before removing
-        obfRmPw( 0 );
+    const bool isServiceAccount = irods::is_service_account();
```
alanking, line 38: We should wrap this in a try-catch block for the `irods::exception`s.

src/iexit.cpp
alanking: Should the message on 42 come before the `unlink` on 40? I realize that would introduce 2 if blocks, just seems upside-down to me.

alanking (commented):
Agreed on the point about parseCmdLineOpt. That function is evil.

Hopefully we won't need to worry about it once we can shift focus to irods_client_cli...

src/iexit.cpp
```diff
@@ -1,19 +1,31 @@
...
+
 void usage( char *prog );
 
+// TODO: replace call to parseCmdLineOpt with something modern and not awful.
+//       make command line arguments more normal and boring:
+//           -y/--yes/--assume-yes: assume yes for all prompts
+//           -f/--force: allow auth file removal for svc acct
+//           -v/--verbose: increase verbosity, can be specified multiple times
+//           -h/--help/-?: show usage
+//       this may require re-implementing some of the functionality of obfRmPw()
```
korydraughn, line 19: Is this required for the issue being addressed?

SwooshyCueb: No, it's more of a note for myself. I'm going to remove it once I've got an issue open for it.

src/iexit.cpp
```diff
@@ -24,14 +36,70 @@ main( int argc, char **argv ) {
...
+        isServiceAccount = irods::is_service_account();
+    }
+    catch (const irods::exception &e) {
+        isServiceAccountError = true;
+        isServiceAccount = true; // to make if/else blocks neater
+        std::fprintf(stderr,
+                     "WARNING: Could not determine whether %s is running as service account "
+                     "due to thrown exception: %s",
+                     argv[0],
+                     e.what());
```
alanking, line 53: Consider using `client_display_what` here to avoid printing the whole stack.

src/iexit.cpp
```diff
@@ -1,19 +1,31 @@
 /*** Copyright (c), The Regents of the University of California            ***
  *** For more information please refer to files in the COPYRIGHT directory ***/
+#include <irods/irods_exception.hpp>
+#include <irods/irods_service_account.hpp>
 #include <irods/rods.h>
 #include <irods/parseCommandLine.h>
 #include <irods/rcMisc.h>
```
alanking, line 7: Since we'll be back here to remove the TODO below anyway, let's remove the Copyright lines and alphabetize these

alanking (commented):
I'm assuming we are expecting one more change to remove the TODO once an issue is created, so I left a couple other notes. Looking real nice

alanking (approved):
Yep, seems good to me. Let's wait for one more approval and I think we can get this in.

src/iexit.cpp
```diff
@@ -24,14 +26,70 @@ main( int argc, char **argv ) {
...
+    if (myRodsArgs.verbose == True) {
+        std::printf("unlink status [%d]\n", status);
+    }
+
+    if (myRodsArgs.verbose == True && !isServiceAccount) {
+        std::printf("Deleting (if it exists) session auth file\n");
+    }
+    if (myRodsArgs.force == True) {
+        if (isServiceAccount) {
+            // MAYBE TODO: try to determine if auth file is safe to remove due to env variables
```
korydraughn, line 59: Leftover? Perhaps this should be captured in an issue.

SwooshyCueb: #6986

korydraughn (commented):
Once the review comment is resolved, please add the pound.

korydraughn (approved):

SwooshyCueb:
Testing requires changing unix user. Not something we can feasibly do with our current test framework.

alanking (approved):

korydraughn:
Cool cool.

PR #385, opened 2023-03-29
[_6627] Output better error messages for invalid path arguments.

d-w-moore:
If the logical path argument given to itree is not an existing collection, we call this an error and return prematurely.


d-w-moore:
Still need to run tests, of course, and perhaps add a test to assert the error outputs.

korydraughn:
I see this is in draft, but is it ready for review?

src/itree.cpp
```diff
@@ -122,6 +122,19 @@ int main(int argc, char** argv){
         auto path = correct_path(vm, env);
         irods::experimental::client_connection conn;
 
+        const auto s = fs::client::status(conn, path);
+
+        if (!fs::client::exists(s))
+        {
+            std::cerr << fmt::format("Error: logical path [{}] does not exist.",path.c_str()) << std::endl;
```
trel, line 129: add space comma?

",path" -> ", path"

src/itree.cpp
```diff
@@ -122,6 +122,19 @@ int main(int argc, char** argv){
...
+            return 1;
+        }
+        else if (fs::client::is_data_object(s))
+        {
+            std::cerr << fmt::format("Error: logical path [{}] refers to a data object, but itree works on collections only.",path.c_str()) << std::endl;
```
trel, line 134: add space comma

",path" -> ", path"

src/itree.cpp
```diff
@@ -122,6 +122,19 @@ int main(int argc, char** argv){
         auto path = correct_path(vm, env);
         irods::experimental::client_connection conn;
 
+        const auto s = fs::client::status(conn, path);
+
+        if (!fs::client::exists(s))
+        {
+            std::cerr << fmt::format("Error: logical path [{}] does not exist.",path.c_str()) << std::endl;
+            return 1;
+        }
```
korydraughn, line 131: This isn't necessary given the more specific existence check below.

src/itree.cpp
```diff
@@ -122,6 +122,19 @@ int main(int argc, char** argv){
...
+        else if (fs::client::is_data_object(s))
+        {
```
korydraughn, line 133: This leaves room for things other than collections to trigger errors.

Instead, you should check for a collection explicitly. For example:
```c++
if (!fs::client::is_collection(s)) {
```
_Notice the opening curly brace has bee moved to the end of the line as well._

If the other if-stmt is removed, this can be rewritten as:
```c++
if (!fs::client::is_collection(conn, path)) {
```
And the call to `fs::client::status(conn, path)` can be removed too.

src/itree.cpp
korydraughn, line 134: Remove `std::endl` and add a `\n` to the format string.

I think `std::cerr` flushes automatically. But if it doesn't, we still avoid the flush due to the use of `std::endl`.

src/itree.cpp
```diff
@@ -122,6 +122,19 @@ int main(int argc, char** argv){
...
         irods::experimental::client_connection conn;
 
+        const auto s = fs::client::status(conn, path);
+
+        if (!fs::client::exists(s))
+        {
+            std::cerr << fmt::format("Error: logical path [{}] does not exist.",path.c_str()) << std::endl;
+            return 1;
+        }
+        else if (fs::client::is_data_object(s))
```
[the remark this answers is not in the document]
d-w-moore: More likely to cover the range of cases without the else. (depending on the meaning of exists).

src/itree.cpp
d-w-moore: yep

src/itree.cpp
d-w-moore: yeps.

src/itree.cpp
d-w-moore: done.

src/itree.cpp
d-w-moore: Yep, done.

korydraughn:
Looks like this can be taken out of draft?

src/itree.cpp
```diff
@@ -122,6 +122,11 @@ int main(int argc, char** argv){
         auto path = correct_path(vm, env);
         irods::experimental::client_connection conn;
 
+        if (!fs::client::is_collection(conn, path)) {
+            std::cerr << fmt::format("Error: logical path [{}] does not refer to a collection.\n", path.c_str());
```
korydraughn, line 126: I don't think we need to report the actual path here. Alternatively, you can remove the `logical path ` from the message since the word, `collection`, implies a logical path.

korydraughn (commented):
Very nice.

src/itree.cpp
d-w-moore: Okey doke, that is done.

src/itree.cpp
```diff
@@ -122,6 +122,11 @@ int main(int argc, char** argv){
         auto path = correct_path(vm, env);
         irods::experimental::client_connection conn;
 
+        if (!fs::client::is_collection(conn, path)) {
+            std::cerr << fmt::format("Error: The specified path does not refer to a collection.\n");
```
alanking, line 126: Don't need the `fmt::format` anymore

d-w-moore: yep, done

korydraughn:
@d-w-moore Anything else left to do for this?

trel:
maybe just a test to see the correct output.   then squash - it's a small enough PR we can see anything new.

d-w-moore:
I think this is it. Will run the tests again and squash.

d-w-moore:
Tests passing for new itree changes, and of course fail without them, which is good.

@korydraughn  - I should the TODO comment as well.  I guess not critical to have the leading-`/`
 test as you pointed out, so I'll remove that assertion too.

korydraughn (approved):
Please squash and pound when ready.

d-w-moore:
> Please squash and pound when ready.

Done!

korydraughn:
You're missing the irods/irods prefix on the issue number.

d-w-moore:
Right, done (final).

alanking:
Will merge along with irods/irods#7007

PR #389, opened 2023-04-06
[irods/irods_6627] Output better error messages for invalid path arguments. (main)

d-w-moore:
cherry-picked from https://github.com/irods/irods_client_icommands/pull/385

If the logical path argument given to itree is not an existing collection, we call this an error and return prematurely.


korydraughn (approved):
Please add the pound.

alanking (approved):

PR #391, opened 2023-04-15
Updated itree help text to include warning about large collections (main)

korydraughn:
Rewrote help text to match look of `itouch` as well.

### DIFF
--- a/src/itree.cpp
+++ b/src/itree.cpp
@@ -82,8 +82,6 @@ int main(int argc, char** argv){
         po::notify(vm);
         if( vm.count("help") ) {
             print_usage();
-            std::cout << desc;
-            printReleaseInfo("itree");
             return 0;
         }
 
@@ -147,7 +145,6 @@ int main(int argc, char** argv){
     }
     catch (const std::exception& e) {
         std::cerr << "Error:" << e.what() << "\n";
-        std::cout << desc << "\n";
         return 1;
     }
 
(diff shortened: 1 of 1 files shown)


trel:
Seems good - consider reporting what a memory exhaustion would look like to the user here - so they could identify if they hit it?

Not sure if that's the best idea...

alanking (commented):
Seems good.

Question: Are we planning to lean on `boost::program_options` (or some equivalent) to generate help text in the future for us as was being done here, or are we going to format the text ourselves? Just trying to understand the school of thought there for the future.

korydraughn:
Let's format the text ourselves. We lose control by relying on Boost. Also, if we ever decide to switch to a different library for handling program options, then we know the help text will be unaffected.

korydraughn:
> Seems good - consider reporting what a memory exhaustion would look like to the user here - so they could identify if they hit it?
> 
> Not sure if that's the best idea...

Did you have something in mind? I'm not sure what that would look like.

src/itree.cpp
```diff
@@ -340,7 +338,42 @@ auto correct_path(const po::variables_map& pm, rodsEnv& env) -> fs::path {
...
+                  The number of spaces used for indenting nested collections.
+                  (defaults to 2).
+  -j, --json      Produce JSON instead of human readable output.
+  -P, --pattern=PATTERN
+                  Filter data objects by a file glob pattern.
+  -p, --pattern-regex=PATTERN
+                  Filter data objects by a regular expression.
+  -x, --regex-extended-syntax
+                  Enable extended syntax for regular expressions.
+  -s, --size      Display the size of each data object.
```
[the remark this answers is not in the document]
korydraughn: Done.

korydraughn:
I've adjusted the help text again to not use "memory exhaustion".

I've also removed the USAGE print out logic when an exception occurs.

Any thoughts/concerns before I squash and pound?

src/itree.cpp
```diff
@@ -340,7 +337,41 @@ auto correct_path(const po::variables_map& pm, rodsEnv& env) -> fs::path {
...
+follow this advice can result in unexpected failures. Consider listing smaller
+collections or using --depth to limit the number of collections traversed.
+
+Options:
+  -c, --collections-only
+                  Only list collections.
+  -C, --color     Print in color (requires ANSI terminal).
+  -f, --fullpath  Print the full path of each item.
+  -F, --classify  Display a forward slash (/) at the end of collections.
+  -i, --ignore-regex=PATTERN
```
trel, line 355: -h help?

korydraughn: Done.

alanking (commented):
Seems good to me. Please squash and leave off #.

korydraughn:
Squashed without pound.

trel (approved):

korydraughn:
Adding pound.

alanking (approved):

PR #395, opened 2023-06-30
[irods 7103] Add relative path support to iticket (main)

Nishant2022:
Update makeFullPath to actually return full path of object or collection

### DIFF
--- a/src/iticket.cpp
+++ b/src/iticket.cpp
@@ -4,11 +4,13 @@
 #include <irods/rcMisc.h>
 #include <irods/rods.h>
 #include <irods/rodsClient.h>
+#include <irods/rodsPath.h>
 
 #include <boost/date_time.hpp>
 
 #include <cstdio>
 #include <locale>
+#include <optional>
 #include <sstream>
 #include <string>
 #include <string_view>
@@ -457,18 +459,18 @@ showTickets( char *inName ) {
     }
 }
 
+auto canonical(const char *inName) -> std::optional<std::string>
+{
+    char out_path[MAX_NAME_LEN]{};
+    if (parseRodsPathStr(inName, &myEnv, out_path) != 0) {
+        return std::nullopt;
+    }
 
-std::string
-makeFullPath( const char *inName ) {
-    std::stringstream fullPathStream;
-    if ( strlen( inName ) == 0 ) {
-        return std::string();
-    }
-    if ( *inName != '/' ) {
-        fullPathStream << cwd << "/";
-    }
-    fullPathStream << inName;
-    return fullPathStream.str();
+    auto* escaped_path = escape_path(out_path);
+    std::optional<std::string> p = escaped_path;
+    std::free(escaped_path);
+
+    return p;
 }
 
 /*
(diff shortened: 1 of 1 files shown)


src/iticket.cpp
```diff
@@ -457,18 +459,17 @@ showTickets( char *inName ) {
...
-
-std::string
-makeFullPath( const char *inName ) {
-    std::stringstream fullPathStream;
-    if ( strlen( inName ) == 0 ) {
-        return std::string();
-    }
-    if ( *inName != '/' ) {
-        fullPathStream << cwd << "/";
+auto makeFullPath(const char *inName) -> std::optional<std::string> {
```
korydraughn, line 462: This looks like the implementation from `istream` and other functions.

Let's just rename this function to match what's in `istream`. That way we know they are related and no one's confused.

Nishant2022:
All unit tests passed. Most core tests passed. the `test_control_plane` test fails on a VM and skips on my local machine. We will look more into that.

Nishant2022:
PR ready for review. All core tests pass apart from `tests_control_plane` (see https://github.com/irods/irods/pull/7153#issuecomment-1618982045). All unit tests pass.

src/iticket.cpp
```diff
@@ -457,18 +459,17 @@ showTickets( char *inName ) {
...
-std::string
-makeFullPath( const char *inName ) {
-    std::stringstream fullPathStream;
-    if ( strlen( inName ) == 0 ) {
-        return std::string();
-    }
-    if ( *inName != '/' ) {
-        fullPathStream << cwd << "/";
+auto canonical(const char *inName) -> std::optional<std::string> {
+    char out_path[MAX_NAME_LEN]{};
```
alanking, line 463:
```suggestion
auto canonical(const char* inName) -> std::optional<std::string>
{
    char out_path[MAX_NAME_LEN]{};
```

alanking (commented):
Looks good. Be sure to fix the commit message for the GitHub linkage, i.e.:

```
[irods/irods#7103] Add relative path support to iticket
```

Nishant2022:
I updated the comment message for GitHub linkage and am ready to add the #'s.

src/iticket.cpp
```diff
@@ -653,8 +655,13 @@ doCommand( char *cmdToken[] ) {
         else {
             makeTicket( myTicket );
         }
-        std::string fullPath = makeFullPath( cmdToken[2] );
-        doTicketOp( "create", myTicket, cmdToken[1], fullPath.c_str(),
+
+        auto fullPath = canonical(cmdToken[2]);
+        if (!fullPath.has_value()) {
+            return -2;
```
korydraughn, line 661: How did we arrive at `-2`?

alanking: My guess is because there's a -1 above this

korydraughn: I had a similar thought, but perhaps @Nishant2022 had a different reason.

Nishant2022: At the top of the current function there is a comment saying:

```c++
/* handle a command,
   return code is 0 if the command was (at least partially) valid,
   -1 for quitting,
   -2 for if invalid
   -3 if empty.
 */
``` 

If the canonical function does not return a value, then the provided path is invalid and the return value should reflect that.

korydraughn: Sounds good to me.

korydraughn: Please mark this as resolved.

korydraughn:
Does the test pass with these changes?

Nishant2022:
All test pass with these changes in addition to my test at https://github.com/irods/irods/pull/7166.

korydraughn (approved):
Excellent. Let's add the pound.

Like @alanking said, don't forget to include the `[irods/irods#7103]` as the commit message prefix.

alanking:
Will merge this along with https://github.com/irods/irods/pull/7166.

PR #407, opened 2023-08-17
[irods/irods#7098] iadmin: Add pam_password config subcommands (main)

alanking:
In service of irods/irods#7098

Companion PR: https://github.com/irods/irods/pull/7245

Will take out of draft when https://github.com/irods/irods/pull/7245 is out of draft.

### DIFF
--- a/src/iadmin.cpp
+++ b/src/iadmin.cpp
@@ -1,6 +1,8 @@
 #include <irods/filesystem.hpp>
+#include <irods/get_grid_configuration_value.h>
 #include <irods/irods_at_scope_exit.hpp>
 #include <irods/irods_client_api_table.hpp>
+#include <irods/irods_configuration_keywords.hpp>
 #include <irods/irods_pack_table.hpp>
 #include <irods/irods_string_tokenize.hpp>
 #include <irods/key_value_proxy.hpp>
@@ -10,9 +12,9 @@
 #include <irods/rodsError.h>
 #include <irods/rodsErrorTable.h>
 #include <irods/rodsLog.h>
-#include <irods/user_administration.hpp>
+#include <irods/set_grid_configuration_value.h>
 #include <irods/user.hpp>
-#include <irods/get_grid_configuration_value.h>
+#include <irods/user_administration.hpp>
 
 #include <algorithm>
 #include <array>
@@ -20,10 +22,11 @@
 #include <iostream>
 #include <string>
 #include <string_view>
-#include <termios.h>
-#include <unistd.h>
 #include <variant>
 #include <vector>
+
+#include <termios.h>
+#include <unistd.h>
 
 #include <boost/lexical_cast.hpp>
 #include <fmt/compile.h>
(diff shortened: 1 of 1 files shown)


src/iadmin.cpp
```diff
@@ -2574,23 +2752,57 @@ usage( char *subOpt ) {
...
+                      rumMsgs,
+                      asqMsgs,
+                      rsqMsgs,
+                      modrepl_usage,
+                      get_delay_server_info_usage,
+                      set_delay_server_usage,
+                      get_pam_password_config_usage,
+                      set_pam_password_config_usage,
+                      helpMsgs,
+                      helpMsgs};
```
alanking, line 2805: Not sure why this is here twice. Will look into it.

I also think I forgot to add my help messages in all the right places.

src/iadmin.cpp
```diff
@@ -2516,6 +2660,40 @@ usage( char *subOpt ) {
...
+        " ",
+        R"_(Example Output:
+
+    {
+        "password_length": 5000,
+        "no_extend": 0,
+        "password_min_time": 3600,
+        "password_max_time": 360000
+    })_",
+        ""};
```
alanking, line 2695: clang-format betrayed me

src/iadmin.cpp
```diff
@@ -2516,6 +2660,40 @@ usage( char *subOpt ) {
         ""
     };
 
+    char* set_pam_password_config_usage[] = {
+        "set_pam_password_config OPTION_NAME OPTION_VALUE",
+        " ",
+        ("Set the specified OPTION_NAME to OPTION_VALUE for pam_password configuration in the local zone in "
+         "R_GRID_CONFIGURATION."),
+        " ",
+        // TODO: Add descriptions for these option names...
```
alanking, line 2669: ...and name them better

src/iadmin.cpp
```diff
@@ -147,6 +149,140 @@ auto get_replica_value(
...
+        std::cerr << "pam_password configuration option name or value is null.\n";
+        return 1;
+    }
+
+    // clang-format off
+    constexpr std::array<const char*, 4> accepted_option_names{
+        "no_extend",
+        "password_length",
+        "password_max_time",
+        "password_min_time"
```
alanking, line 245: Maybe use keywords for these

src/iadmin.cpp
```diff
@@ -147,6 +149,140 @@ auto get_replica_value(
...
+
+    if (!option_name_is_valid) {
+        std::cerr << fmt::format("Invalid pam_password configuration option [{}]\n", _option_name);
+        return 1;
+    }
+
+    GridConfigurationInput input{};
+
+    if (std::strlen(_option_value) >= sizeof(input.option_value)) {
+        std::cerr << fmt::format("Invalid pam_password configuration value: Must be shorter than 2700 characters.\n");
```
trel, line 262: 2700 hard coded here...  probably never going to change, but could be 'derived', same as the check in the line above...

alanking: Yep, will do

src/iadmin.cpp
alanking: "help" and "h" are recognized by `iadmin`, which is why there are two of these.

trel: well, feels weird, but explained!

src/iadmin.cpp
korydraughn: Lol, disable clang-format for that part.

src/iadmin.cpp
```diff
@@ -147,6 +149,140 @@ auto get_replica_value(
...
+        rodsLogError(LOG_ERROR, SYS_UNKNOWN_ERROR, e.what());
+        return false;
+    }
+
+    return true;
+} // check_that_connected_user_is_rodsadmin
+
+auto print_pam_password_config() -> int
+{
+    if (!check_that_connected_user_is_rodsadmin(Conn)) {
```
korydraughn, line 189: Consider renaming so that usage reads like a question.

- `is_connected_as_rodsadmin`
- `is_rodadmin`
- `is_privileged_user`
- `connected_user_is_rodsadmin`

Those are just some examples. Feel free to come up with something else.

src/iadmin.cpp
```diff
@@ -147,6 +149,140 @@ auto get_replica_value(
     return v;
 } // get_replica_value
 
+auto check_that_connected_user_is_rodsadmin(RcComm* _comm) -> bool
```
korydraughn, line 152: `RcComm&`?

src/iadmin.cpp
```diff
@@ -147,6 +149,140 @@ auto get_replica_value(
...
+auto set_pam_password_config(const char* _option_name, const char* _option_value) -> int
+{
+    if (!check_that_connected_user_is_rodsadmin(Conn)) {
+        return 1;
+    }
+
+    if (!_option_name || !_option_value) {
+        std::cerr << "pam_password configuration option name or value is null.\n";
+        return 1;
+    }
```
korydraughn, line 238: Consider checking the options first to avoid network traffic.

src/iadmin.cpp
```diff
@@ -2516,6 +2660,40 @@ usage( char *subOpt ) {
...
+        " ",
+        ("Set the specified OPTION_NAME to OPTION_VALUE for pam_password configuration in the local zone in "
+         "R_GRID_CONFIGURATION."),
+        " ",
+        // TODO: Add descriptions for these option names...
+        "The list of accepted OPTION_NAMEs is:",
+        " - password_length",
+        " - no_extend",
+        " - password_min_time",
+        " - password_max_time",
```
korydraughn, line 2674: Should we sort these? Hmm ...

src/iadmin.cpp
```diff
@@ -2516,6 +2660,40 @@ usage( char *subOpt ) {
...
+        ""};
+
+    char* get_pam_password_config_usage[] = {
+        "get_pam_password_config",
+        " ",
+        "Prints pam_password configuration from the local zone grid configuration as JSON.",
+        " ",
+        ("This command allows administrators to view pam_password configuration values without running queries "
+         "directly against the database."),
```
korydraughn, line 2683: Why parentheses here?
Wouldn't this result in the line being really long in the terminal?

alanking: Yep. The compiler gives a warning here because if you split up the string it isn't sure if you meant to put a comma at the end of the line. I'm happy to just split this into two lines in the output if that's preferable.

src/iadmin.cpp
alanking: The idea behind the name was to hint at side effects (the function will print messages for various errors that occur) and not give the impression that it's a pure function. I see that it did not do its job well :p 

If we're okay with the side effects, I'll just rename it to `connected_user_is_rodsadmin` which is what I had originally. Otherwise, I'll think about a better approach here (to me, top candidate is just not using a function).

src/iadmin.cpp
alanking: Sure

src/iadmin.cpp
alanking: Will do

src/iadmin.cpp
korydraughn: Don't we want these lines to be split via a comma so the help text honors the imaginary margin shared by all lines?

src/iadmin.cpp
alanking: The side effect of the function which is only supposed to return whether the connected user is a rodsadmin is that it also prints messages for various errors that occur.

At least in my mind, a "clean" implementation of this function would be to only return whether the connected user is a rodsadmin and leave it up to the caller to print errors and whatnot. That's all I was getting at there. Might not be a big deal that the function prints things - I mostly did this because I didn't want to see all that mess in `print_pam_password_config`

src/iadmin.cpp
alanking: Seeing as how we will be updating the docs anyway, we can sort them in all the places

src/iadmin.cpp
alanking: Actually I think this is right. It at least looks like the `get_delay_server_info` help text, so I'm just going to leave it.

src/iadmin.cpp
alanking: I've decided to just remove this function. Some code duplication now, but at least there's no hidden surprises :)

src/iadmin.cpp
trel: let's sort.   it's the most defensible position.

alanking:
irods/irods#7245 has changed purpose. This will be updated soon to be in service of irods/irods#7274

src/iadmin.cpp
alanking: This function has been removed. Resolving.

src/iadmin.cpp
alanking: This function has been removed. Resolving.

src/iadmin.cpp
alanking: Done

src/iadmin.cpp
alanking: Done

src/iadmin.cpp
alanking: Done, I think

korydraughn:
What are next steps for this?

alanking:
Preparing companion PR in irods/irods... Will notify when this is ready for review.

alanking:
Ready for review

src/iadmin.cpp
```diff
@@ -2516,6 +2669,24 @@ usage( char *subOpt ) {
...
+        "set_grid_configuration NAMESPACE OPTION_NAME OPTION_VALUE",
+        " ",
+        "Set the specified OPTION_NAME to OPTION_VALUE for the specified NAMESPACE in the local zone in",
+        "R_GRID_CONFIGURATION. If NAMESPACE or OPTION_NAME is not valid, an error is returned.",
+        ""};
+
+    char* get_grid_configuration_usage[] = {
+        "get_grid_configuration NAMESPACE OPTION_NAME",
+        " ",
+        "Prints local zone grid configuration value for specified namespace and option name.",
```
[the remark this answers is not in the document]
alanking: Done

src/iadmin.cpp
```diff
@@ -2516,6 +2669,24 @@ usage( char *subOpt ) {
...
+        " ",
+        "This command allows administrators to view grid configuration values without running queries directly",
+        "against the database. If NAMESPACE or OPTION_NAME is not valid, an error is returned.",
```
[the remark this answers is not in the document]
alanking: Done

korydraughn (approved):
Looks good to me.

Squash, pound, and merge when ready.

alanking:
I'm gonna hold off and merge this at the same time as irods/irods#7331

PR #408, opened 2023-08-20
Address Sanitizer fixes (main)

korydraughn:
Companion PR to irods/irods#7269.

### DIFF
--- a/src/ilsresc.cpp
+++ b/src/ilsresc.cpp
@@ -1,8 +1,10 @@
+#include <irods/irods_at_scope_exit.hpp>
 #include <irods/irods_children_parser.hpp>
 #include <irods/irods_client_api_table.hpp>
 #include <irods/irods_exception.hpp>
 #include <irods/irods_pack_table.hpp>
 #include <irods/irods_resource_constants.hpp>
+#include <irods/rcMisc.h>
 #include <irods/rods.h>
 #include <irods/rodsClient.h>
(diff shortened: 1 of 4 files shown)


src/ilsresc.cpp
```diff
@@ -490,6 +497,8 @@ main( int argc, char **argv ) {
     rcComm_t *Conn = rcConnect( myEnv.rodsHost, myEnv.rodsPort, myEnv.rodsUserName,
                       myEnv.rodsZone, 0, &errMsg );
 
+    const auto disconnect = irods::at_scope_exit{[&Conn] { rcDisconnect(Conn); }};
```
alanking, line 500: If `Conn` is `NULL`, we do not need to call `rcDisconnect`. Why move it here?

alanking (commented):
Just one thing, pretty minor

src/ilsresc.cpp
alanking: Got it. Thanks

korydraughn: I'm going to leave this as is since it's what was tested.

We can always adjust later.

alanking (commented):
Please add issue numbers and squash (or don't) to taste

korydraughn:
Attached issue number 6634 to commits.

Are we ready for pounds?

alanking (approved):
\# it

korydraughn:
#'d

PR #420, opened 2023-10-25
[irods/irods_6224] remove unused test mode from irule

trel:
have not tested yet


alanking (approved):
Big minus. Please # when confirmed/tested.

A quick grep suggests (as is noted in the issue) that "looptest" is not referenced anywhere in irods/irods.

trel:
built, installed, and help text appears correct.

#'d.

PR #421, opened 2023-10-25
[irods/irods_6223] add inputFile size limit to irule help text (4-3-stable)

trel:
and some other rewording from 'line' to 'part'


alanking (approved):
Seems good to me. Maybe we get thoughts from one more person on this one before we put it in...

trel:
built and 'tested' via `irule -h` and looking at the words.

korydraughn (approved):
Reads fine.

I think what would help this documentation is actual examples of the different forms of input.

That can come later though. The objective of this PR is satisfied IMO.

trel:
I see - we have a number of them in the examples linked... but yes, I agree.

trel:
#'d.

PR #424, opened 2023-11-15
[irods/irods#7421] irm: expand docs force option (main)

stsnel:
Note that this option makes irm ignore non-existent data objects.


korydraughn:
Let's change the issue number to irods/irods#7421 here as well.

alanking (approved):

PR #425, opened 2023-11-15
[irods/irods#7421] irm: expand docs force option (4-3-stable)

stsnel:
Note that this option makes irm ignore non-existent data objects.

### DIFF
--- a/src/irm.cpp
+++ b/src/irm.cpp
@@ -122,6 +122,7 @@ usage() {
         " ",
         "Options are:",
         " -f  force - Immediate removal of data objects without putting them in trash.",
+        "             Ignores non-existent data objects and collections.",
         " -r  recursive - Recursively remove the target collection, all data objects in the ",
         "                 collection, and all subcollections.",
         " -v  verbose",


src/irm.cpp
```diff
@@ -122,6 +122,8 @@ usage() {
         " ",
         "Options are:",
         " -f  force - Immediate removal of data objects without putting them in trash.",
+        "             The command does not show an error message or return a non-zero ",
+        "             status code for non-existent data objects.",
```
korydraughn, line 126: This is slightly incorrect (based on the implementation).

How about ...
```
Ignores non-existent data objects and collections.
```

korydraughn (commented):
Thanks for the contribution.

Just one tweak and we'll get this added.

src/irm.cpp
stsnel: Thank you, that is indeed better. I've changed the text.

korydraughn:
The issue number linked cannot be used because it's already part of a closed milestone.

I've created irods/irods#7421 for this work. Please use that.

alanking (approved):

PR #429, opened 2024-01-04
[irods/irods2600] change copies to replicas in help text (master)

trel:


src/itrim.cpp
```diff
@@ -106,10 +106,9 @@ usage() {
         " ",
         "Reduce the number of replicas of a dataObject in iRODS by deleting some replicas.",
         "Nothing will be done if this is less than or equal to numCopies. The -n and",
-        "the -S options are used to specified which copies to delete. If neither of",
+        "the -S options are used to specify which replica to delete. If neither of",
```
korydraughn, line 109: -S can target multiple replicas. Perhaps `replica` should be plural?

src/itrim.cpp
```diff
@@ -122,17 +121,17 @@ usage() {
...
         " -M  admin - admin user uses this option to trim other users files",
         " -n  replNum - the replica number of the replica to be deleted",
-        " -N  numCopies - minimum number of most current copies to keep. The default",
+        " -N  numCopies - minimum number of good replicas to keep. The default",
         "     value is 2.",
         " -r  recursive - trim the whole subtree",
         " -S  srcResource - specifies the resource of the replica to be deleted.",
-        "     If specified, only copies stored in this resource will be candidate",
+        "     If specified, only replicas stored in this resource will be candidate",
         "     for the deletion.",
```
korydraughn, line 129: The first line seems weird. It's worded as if the option is used to target a **single** replica, yet your updated line mentions **replicas** (which is correct).

korydraughn (commented):
Looks good.

src/itrim.cpp
trel: agreed, will update.

src/itrim.cpp
trel: agreed, will update.

trel: done.

korydraughn (approved):
Squash/Pound it.

trel:
#'d

PR #447, opened 2024-03-11
Changed iadmin rmgroup to use new api convention

FifthPotato:
Needed for for https://github.com/irods/irods/pull/7533
Tests will fail without this change.

### DIFF
--- a/src/iadmin.cpp
+++ b/src/iadmin.cpp
@@ -15,6 +15,7 @@
 #include <irods/set_grid_configuration_value.h>
 #include <irods/user.hpp>
 #include <irods/user_administration.hpp>
+#include <irods/version.hpp>
 
 #include <algorithm>
 #include <array>
@@ -1624,8 +1625,9 @@ doCommand( char *cmdToken[], rodsArguments_t* _rodsArgs = 0 ) {
         return 0;
     }
     if ( strcmp( cmdToken[0], "rmgroup" ) == 0 ) {
-        generalAdmin( 0, "rm", "user", cmdToken[1],
-                      myEnv.rodsZone, "", "", "", "", "", "" );
+        const auto version = irods::to_version(Conn->svrVersion->relVersion);
+        const char* entity_type = (version && version.value() > irods::version{4, 3, 1}) ? "group" : "user";
+        generalAdmin(0, "rm", const_cast<char*>(entity_type), cmdToken[1], myEnv.rodsZone, "", "", "", "", "", "");
         return 0;
     }
 
--- a/.clang-format
+++ b/.clang-format
@@ -1,0 +1,203 @@
+---
+DisableFormat:   false
+
+Language:        Cpp
+Standard:        c++20
+
+ColumnLimit:     120
+UseCRLF:         false
+
+#UseTab:          AlignWithSpaces
+UseTab:          Never
+TabWidth:        4
+IndentWidth:     4
+ConstructorInitializerIndentWidth: 4
+
+IndentPPDirectives: AfterHash
+PPIndentWidth:   2
+
+IndentAccessModifiers: false
+AccessModifierOffset: -2
+
+AlignAfterOpenBracket: Align
+#AlignAfterOpenBracket: BlockIndent
+AlignArrayOfStructures: None
+AlignConsecutiveAssignments: None
+AlignConsecutiveBitFields: Consecutive
+AlignConsecutiveDeclarations: None
+AlignConsecutiveMacros: AcrossEmptyLinesAndComments
+AlignEscapedNewlines: Left
+AlignOperands:   Align
+AlignTrailingComments: false
+
+AllowAllArgumentsOnNextLine: true
+AllowAllParametersOfDeclarationOnNextLine: false
+AllowAllConstructorInitializersOnNextLine: false
+ConstructorInitializerAllOnOneLineOrOnePerLine: false
+
+AllowShortBlocksOnASingleLine: Empty
+AllowShortCaseLabelsOnASingleLine: false
+AllowShortEnumsOnASingleLine: false
+AllowShortFunctionsOnASingleLine: None
+AllowShortIfStatementsOnASingleLine: Never
+AllowShortLambdasOnASingleLine: All
+AllowShortLoopsOnASingleLine: false
+
+IndentCaseBlocks: false
+IndentCaseLabels: true
+IndentExternBlock: AfterExternBlock
+IndentGotoLabels: true
+IndentRequires:  true
+#IndentRequiresClause: true
+IndentWrappedFunctionNames: false
+LambdaBodyIndentation: Signature
+NamespaceIndentation: All
+
+AlwaysBreakAfterReturnType: None
+AlwaysBreakAfterDefinitionReturnType: None
+AlwaysBreakBeforeMultilineStrings: false
+AlwaysBreakTemplateDeclarations: Yes
+
+BinPackArguments: false
+BinPackParameters: false
+
+BreakBeforeBinaryOperators: None
+BreakBeforeConceptDeclarations: true
+BreakBeforeInheritanceComma: true
+BreakBeforeTernaryOperators: true
+BreakConstructorInitializers: BeforeComma
+BreakInheritanceList: BeforeComma
+BreakStringLiterals: true
+BreakBeforeBraces: Custom
+BraceWrapping:
+   AfterCaseLabel:  false
+   AfterClass:      true
+   AfterControlStatement: MultiLine
+   AfterEnum:       true
+   AfterExternBlock: false
+   AfterFunction:   true
+   AfterNamespace:  true
+   AfterStruct:     true
+   AfterUnion:      true
+   BeforeCatch:     true
+   BeforeElse:      true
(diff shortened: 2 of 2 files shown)


src/iadmin.cpp
```diff
@@ -1624,7 +1624,7 @@ doCommand( char *cmdToken[], rodsArguments_t* _rodsArgs = 0 ) {
         return 0;
     }
     if ( strcmp( cmdToken[0], "rmgroup" ) == 0 ) {
-        generalAdmin( 0, "rm", "user", cmdToken[1],
+        generalAdmin( 0, "rm", "group", cmdToken[1],
```
trel, line 1627: but... only when talking to a server that is 4.3.2 or greater?

alanking: We have not historically considered the server version (as far as I know) as iCommands versions have always been tied to a particular server version.

Are we going to start spanning server versions in iCommands version 4.3.2? If not, I think the change as-is is fine?

trel: oh, i see - not a PRC sniff-the-server situation...   hmm...

let's think on that...

FifthPotato: Is there something I need to add for this?

trel: What does a 4.3.1 and earlier server do when it gets this request from 4.3.2+ icommands?

FifthPotato: The code seems to imply it'll just throw CAT_INVALID_ARGUMENT, but I'll rebuild a 4.3.1 and check.

FifthPotato: Confirmed that's what happens, CAT_INVALID_ARGUMENT.

trel: so, 4.3.2 icommands can't do this with a 4.3.1 server.

my first reaction is that we should sniff and send compatible request.

src/iadmin.cpp
```diff
@@ -1624,8 +1625,14 @@ doCommand( char *cmdToken[], rodsArguments_t* _rodsArgs = 0 ) {
...
-        generalAdmin( 0, "rm", "user", cmdToken[1],
-                      myEnv.rodsZone, "", "", "", "", "", "" );
+        if(irods::to_version(Conn->svrVersion->relVersion) > irods::version{4, 3, 1}) {
+            generalAdmin( 0, "rm", "group", cmdToken[1],
+                          myEnv.rodsZone, "", "", "", "", "", "" );
+        }
+        else {
+            generalAdmin( 0, "rm", "user", cmdToken[1],
+                          myEnv.rodsZone, "", "", "", "", "", "" );
+        }
```
korydraughn, line 1635: Consider isolating only the things that are changing to the if-ladder.

For example:
```cpp
const char* entity_type = (/* version > 4.3.1 */) ? "group" : "user";
generalAdmin(0, "rm", entity_type, ...);
```

With your current implementation, it's possible for the calls to desync.

src/iadmin.cpp
```diff
@@ -1624,8 +1625,14 @@ doCommand( char *cmdToken[], rodsArguments_t* _rodsArgs = 0 ) {
         return 0;
     }
     if ( strcmp( cmdToken[0], "rmgroup" ) == 0 ) {
-        generalAdmin( 0, "rm", "user", cmdToken[1],
-                      myEnv.rodsZone, "", "", "", "", "", "" );
+        if(irods::to_version(Conn->svrVersion->relVersion) > irods::version{4, 3, 1}) {
```
korydraughn, line 1628: Doesn't `to_version` return an optional?

Also, avoid implicit conversions.

src/iadmin.cpp
```diff
@@ -1624,9 +1625,13 @@ doCommand( char *cmdToken[], rodsArguments_t* _rodsArgs = 0 ) {
...
     }
     if ( strcmp( cmdToken[0], "rmgroup" ) == 0 ) {
-        generalAdmin( 0, "rm", "user", cmdToken[1],
-                      myEnv.rodsZone, "", "", "", "", "", "" );
-        return 0;
+      const auto version = irods::to_version(Conn->svrVersion->relVersion);
+      const char *entity_type =
+          (version && version.value() > irods::version{4, 3, 1}) ? "group"
+                                                                 : "user";
+      generalAdmin(0, "rm", const_cast<char*>(entity_type), cmdToken[1], myEnv.rodsZone, "", "",
```
korydraughn, line 1632: The need for `const_cast` sucks, but I get it. I can never remember if casting away const is UB. I bet it is.

To get around it, allocate a buffer on the stack, copy the string into it, and pass the buffer to generalAdmin.

Alternatively, assign the string to a non-const std::string and use `.data()` to get at the underlying buffer.

FifthPotato: Hmm, a bit of research turns up that it's only UB if you try to modify it through the un-const'ed pointer, I think? Would it be legal in this case, then, since I don't suppose `generalAdmin` is going to modify its 3rd argument anyhow (since string literals were passed in its place before...)
I had originally wanted to do this because I wanted to avoid the extra copy/alloc if it wasn't necessary.
Or do you think it's just not worth the trouble and I should just do the copy/alloc?

korydraughn: That's fair. I've used `const_cast` in code under the assumption that nothing will modify it.

Letting it ride.

korydraughn (commented):
Changes look good.

What's left? Testing?

korydraughn:
There's still one review comment that needs to be resolved _(seems the work is already done)_?

FifthPotato:
> Changes look good.
> 
> What's left? Testing?

Yep, just gotta run it against everything, then both this and https://github.com/irods/irods/pull/7533 should be ready together. I'll update this when tests complete.

korydraughn:
Sounds good.

FifthPotato:
All tests pass! Had only 2 failures on first pass, then none after rerunning those.

alanking:
Which tests failed? Please note them here if you still have access to that information.

FifthPotato:
> Which tests failed? Please note them here if you still have access to that information.

They were test_delay_queue.Test_Execution_Frequency test_dynamic_peps

FifthPotato:
> Good.
> 
> Looks like the formatting is off. Please correct that.

I think I reindented good, now?

korydraughn:
Do you feel those test failures are related to your changes?
Did you inspect the XML test output for anything weird?

> I think I reindented good, now?

Looking now.

src/iadmin.cpp
```diff
@@ -1624,9 +1625,13 @@ doCommand( char *cmdToken[], rodsArguments_t* _rodsArgs = 0 ) {
...
-                      myEnv.rodsZone, "", "", "", "", "", "" );
-        return 0;
+      const auto version = irods::to_version(Conn->svrVersion->relVersion);
+      const char *entity_type =
+          (version && version.value() > irods::version{4, 3, 1}) ? "group"
+                                                                 : "user";
+      generalAdmin(0, "rm", const_cast<char *>(entity_type), cmdToken[1],
+                   myEnv.rodsZone, "", "", "", "", "", "");
+      return 0;
     }
```
korydraughn, line 1635: Indentation is still not quite right.

FifthPotato:
> Do you feel those test failures are related to your changes? Did you inspect the XML test output for anything weird?
> 
> > I think I reindented good, now?
> 
> Looking now.

I didn't inspect, but I ran the two failed tests right after and they passed without issues. Also, indented properly now.

korydraughn:
Good. Please squash to taste, no #'s

korydraughn:
Also, please copy the `.clang-format` file from irods/irods into this repo and make that the first commit of this PR. You can then format the second commit (which includes the real meat of this PR) according to that.

In summary, we should see a commit history like the following:
1. `[?] Add .clang-format`
2. `[irods/irods ?] Changed iadmin rmgroup to use new API argument`

Then you'd run clang-format on your changes.

Make sense?

FifthPotato:
> Also, please copy the `.clang-format` file from irods/irods into this repo and make that the first commit of this PR. You can then format the second commit (which includes the real meat of this PR) according to that.
> 
> In summary, we should see a commit history like the following:
> 
>     1. `[?] Add .clang-format`
> 
>     2. `[irods/irods ?] Changed iadmin rmgroup to use new API argument`
> 
> 
> Then you'd run clang-format on your changes.
> 
> Make sense?

I think so, let me give that a go...

FifthPotato:
Squished as requested, I think?

korydraughn:
Yep, looks good.

Open an issue in irods/irods for adding the `.clang-format` file to this project and attach it to the clang-format commit.

Let us know once that's complete.

FifthPotato:
I believe I have done the thing.

alanking (approved):
Seems right to me. Deferring to @korydraughn

korydraughn (approved):
Yep. Pound it!

FifthPotato:
Octothorpe'd

alanking:
@FifthPotato - Please cherry-pick to 4-3-stable. Thanks

korydraughn:
@FifthPotato No #'s on the cherry-picks yet.

PR #453, opened 2024-03-18
GenQuery2 - Add new icommand: iquery (main)

korydraughn:
Depends on irods/irods#7582.

This works as intended.

Whether we want to include a new icommand or add this functionality to `iquest` is up for debate.


trel:
oh, right - i was not assuming we'd have a new icommand.

iquery was to allow installation *alongside* iquest while this was a plugin/experiment.

i think for the actual absorption, we just want an option in iquest.

trel:
but also... this has different options itself... so...

src/iquery.cpp
```diff
@@ -0,0 +1,178 @@
...
+        const auto ec =
+                procApiRequest(static_cast<RcComm*>(conn),
+                           IRODS_APN_GENQUERY2,
+                           &input,
+                           nullptr,
+                           reinterpret_cast<void**>(&sql), // NOLINT(cppcoreguidelines-pro-type-reinterpret-cast)
+                           nullptr);
+#else
+        const auto ec = rc_genquery2(static_cast<RcComm*>(conn), &input, &sql);
+#endif
```
korydraughn, line 109: Remove preprocessor conditionals.

src/iquery.cpp
```diff
@@ -0,0 +1,178 @@
...
+#include <algorithm>
+#include <cstdlib>
+#include <iostream>
+
+auto print_usage_info() -> void;
+auto print_columns_info() -> void;
+
+int main(int _argc, char* _argv[]) // NOLINT(modernize-use-trailing-return-type)
+{
+    set_ips_display_name("iquery (experimental)");
```
korydraughn, line 24: Should we remove `(experimental)`?

src/iquery.cpp
```diff
@@ -0,0 +1,178 @@
...
+Options:
+  -c, --columns         List columns supported by GenQuery2.
+      --sql-only        Print the SQL generated by the parser. The generated
+                        SQL will not be executed.
+  -z, --zone=ZONE_NAME  The name of the zone to run the query against. Defaults
+                        to the local zone.
+  -h, --help            Display this help message and exit.
+)_");
+
+	char name[] = "iquery (experimental)";
```
korydraughn, line 157: Should we remove `(experimental)`?

src/iquery.cpp
trel: i think so - it's been proven/tested in the field and is the future.

src/iquery.cpp
trel: same

i think so - it's been proven/tested in the field and is the future.

korydraughn:
After some discussion, we've decided to leave `iquest` as is and continue with introducing a new icommand.

src/iquery.cpp
korydraughn: Done

src/iquery.cpp
korydraughn: Done

src/iquery.cpp
korydraughn: Done

src/iquery.cpp
```diff
@@ -0,0 +1,167 @@
...
+
+QUERY_STRING is expected to be a string matching the GenQuery2 syntax. Failing
+to meet this requirement will result in an error.
+
+If QUERY_STRING is a hyphen (-) or is empty, input is read from stdin. Input
+taken via stdin will be viewed as the QUERY_STRING to execute. For example:
+
+    echo select COLL_NAME, DATA_NAME | iquery
+
+Mandatory arguments to long options are mandatory for short options too.
```
trel, line 135: looking at the options... what does this mean for this icommand?

can this line be removed?

trel (commented):
looks pretty good

we have an issue for this?   just using the irods/irods issue?

src/iquery.cpp
korydraughn: That line is primarily about the `-z` option.

It was modeled after other POSIX commands. The man page for `ls` uses those exact words.

It's not required, but it was included so that it's clear to users that `-z` takes the same argument as `--zone`.

trel: right, `zone` is the only option that takes a parameter.

the use of 'mandatory' and that it has a default value is what threw me off a little.  but it is not confusing, just seemed ... superfluous.

definitely fine to leave it as is.

korydraughn: Cool. Leaving as is.

src/iquery.cpp
```diff
@@ -0,0 +1,173 @@
...
+        auto c = std::begin(items);
+
+        if (c.key().empty()) {
+            fmt::print("{:{}} (derived)\n", _v.key(), w);
+            return;
+        }
+
+        fmt::print("{:{}} ({}.{})\n", _v.key(), w, c.key(), c.value().template get_ref<const std::string&>());
+    });
+} // print_columns_info
```
korydraughn, line 173: s_info -> _mappings

korydraughn: Done.

src/iquery.cpp
```diff
@@ -0,0 +1,173 @@
...
+    return 1;
+} // main
+
+auto print_usage_info() -> void
+{
+    fmt::print(R"_(iquery - Query the iRODS Catalog
+
+Usage: iquery [OPTION]... QUERY_STRING
+
+Queries the iRODS Catalog using GenQuery2.
```
alanking, line 126: Is this iCommand considered experimental given that the API it calls is experimental? Should we mention that in the help text?

src/iquery.cpp
```diff
@@ -0,0 +1,173 @@
...
+            input.zone = zone.data();
+        }
+
+        if (vm["sql-only"].as<bool>()) {
+            input.sql_only = 1;
+        }
+
+        irods::experimental::client_connection conn;
+        char* output{};
+        irods::at_scope_exit free_sql{[&output] {
```
alanking, line 89:
```suggestion
        irods::at_scope_exit free_output{[&output] {
```

alanking (commented):
Seems good overall. Just had a couple of suggestions.

src/iquery.cpp
korydraughn: The help text originally included `(experimental)` beside the icommand name.

Reverting that change makes sense given the actual API could change. Although one could argue the interface of `iquery` is independent of the underlying API.

alanking: Fair enough. I think it's fine as-is, then

korydraughn: The safest thing to do is to revert the removal of the `(experimental)` tag.

Will think about that a little.

korydraughn: Decided to restore the experimental tag to indicate to users that things can change in the future.

korydraughn:
#'d

alanking (approved):

PR #456, opened 2024-03-22
[irods/irods#6953] irepl: Deprecate "backupMode" (main)

alanking:
Addresses irods/irods#6953

Companion PR: https://github.com/irods/irods/pull/7591

### DIFF
--- a/src/irepl.cpp
+++ b/src/irepl.cpp
@@ -160,7 +160,7 @@ usage() {
         " ",
         "Options are:",
         " -a  all - if used with -U, update all stale copies",
-        " -B  Backup mode - if a good copy already exists in this",
+        " -B  Backup mode - [Deprecated] if a good copy already exists in this",
         "     resource, don't make another copy.",
         " -P  output the progress of the replication.",
         " -Q  use RBUDP (datagram) protocol for the data transfer",


src/irepl.cpp
```diff
@@ -160,7 +164,7 @@ usage() {
         " ",
         "Options are:",
         " -a  all - if used with -U, update all stale copies",
-        " -B  Backup mode - if a good copy already exists in this",
+        " -B  Backup mode - (Deprecated) if a good copy already exists in this",
```
trel, line 167: hesitate to suggest this - but consider changing the word 'copy' on this line and the next to be 'replica'... but ... not even sure that is correct/true.

ignore with impunity.

alanking: I think... I'm going to leave this as-is. Copy might actually be the right word in this case - I simply do not know.

trel: Defer.  Resolving.

src/irepl.cpp
```diff
@@ -96,6 +96,10 @@ main( int argc, char **argv ) {
         exit( 7 );
     }
 
+    if (0 != myRodsArgs.backupMode) {
+        printf("Warning: Backup Mode (-B) is deprecated. Its use should be avoided.\n");
```
MartinFlores751, line 100: Can `std::printf(...)`.

MartinFlores751 (commented):
Bonus item.

src/irepl.cpp
alanking: Done

src/irepl.cpp
```diff
@@ -96,6 +96,10 @@ main( int argc, char **argv ) {
         exit( 7 );
     }
 
+    if (0 != myRodsArgs.backupMode) {
+        std::printf("Warning: Backup Mode (-B) is deprecated. Its use should be avoided.\n");
+    }
```
korydraughn, line 101: Do we print out messages about flags being deprecated for other icommands and options?

alanking: Yes.

https://github.com/irods/irods/issues/4860
https://github.com/irods/irods/issues/7498

alanking:
Removed warning message

korydraughn (approved):
Squash/Pound when ready.

alanking:
Squashed. Will # along with https://github.com/irods/irods/pull/7591

korydraughn (approved):
Pound when ready.

alanking:
Done. Mergin...

PR #459, opened 2024-03-27
[irods/irods_6363] update quota helptext (main)

trel:


alanking (approved):
Looks like it clears it up to me. Will defer to @korydraughn for this one.

korydraughn (commented):
The wording is much better, but the `ResourceName-or-'total'` parameter still needs to change. Seeing that in help text is just not common.

What if the arg list became something like ...
```
suq User Target Value
```
My personal preference would be to mirror the man pages and capitalize the args.
```
suq USER TARGET VALUE
``` 
Then you could replace things like `first parameter` with `USER parameter` making it even more clear.

trel:
i like it.

trel:
updated, but didn't go all caps as no other iadmin subcommand uses that formatting.

src/iadmin.cpp
```diff
@@ -2561,35 +2561,35 @@ usage( char *subOpt ) {
...
     };
 
     char *sgqMsgs[] = {
-        " sgq Group ResourceName-or-'total' Value (set group quota)",
+        " sgq Group Target Value (set group quota)",
         " ",
         "Set a storage quota for a group.",
         " ",
-        "The first parameter to this sub-command is the name of the iRODS group.",
+        "The User parameter is the name of the iRODS group.",
```
korydraughn: User -> Group?

trel: had already caught it - next commit.

korydraughn (commented):
Looks like it's ready for squashing.

trel:
squashed

korydraughn (approved):
Pound it.

trel:
#'d

PR #426, opened 2023-11-28
[irods/irods#7376] imeta: add set_ips_display_name call (main)

mstfdkmn:
This will display the command name alongside the connection information (spOption).

PR #467, opened 2024-04-16
[irods/irods#7470] Ensure icommands set SP_OPTION (main)

MartinFlores751:
Ensure that all icommands set `SP_OPTION`. Used `set_ips_display_name(...)` for additional warning log if an error occurs.

### DIFF
--- a/src/ierror.cpp
+++ b/src/ierror.cpp
@@ -8,6 +8,7 @@ void usage( char *prog );
 
 int
 main( int argc, char **argv ) {
+    set_ips_display_name("ierror");
 
     signal( SIGPIPE, SIG_IGN );
 
--- a/src/iget.cpp
+++ b/src/iget.cpp
@@ -10,6 +10,7 @@
 #include <irods/rodsPath.h>
 #include <irods/getUtil.h>
 #include <irods/rcGlobalExtern.h>
+#include <irods/rcMisc.h>
 #include <irods/irods_client_api_table.hpp>
 #include <irods/irods_pack_table.hpp>
 #include <irods/irods_parse_command_line_options.hpp>
@@ -20,6 +21,7 @@ void usage( FILE* );
 
 int
 main( int argc, char **argv ) {
+    set_ips_display_name("iget");
 
     signal( SIGPIPE, SIG_IGN );
 
--- a/src/iput.cpp
+++ b/src/iput.cpp
@@ -10,6 +10,7 @@
 #include <irods/rodsPath.h>
 #include <irods/putUtil.h>
 #include <irods/rcGlobalExtern.h>
+#include <irods/rcMisc.h>
 #include <irods/irods_client_api_table.hpp>
 #include <irods/irods_pack_table.hpp>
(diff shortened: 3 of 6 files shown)


src/ierror.cpp
```diff
@@ -8,6 +8,7 @@ void usage( char *prog );
 
 int
 main( int argc, char **argv ) {
+    set_ips_display_name("ierror");
```
MartinFlores751, line 11: Goes for all calls to `set_ips_display_name(...)` in this PR, do we prefer hard-coding the name or using another method to get the command name?

korydraughn: Hard coding is fine. The names of these commands aren't going to change.

korydraughn (commented):
Have you confirmed the names appear in the output of `ips`?

src/imeta.cpp
```diff
@@ -2136,6 +2136,7 @@ int display_help(const std::vector<std::string>& _sub_args) {
 
 int main( int argc, const char **argv )
 {
+    set_ips_display_name("imeta");
     signal( SIGPIPE, SIG_IGN );
 
     utils::set_ips_display_name(boost::filesystem::path{argv[0]}.filename().c_str());
```
alanking, line 2142: Hmm... what's going on here?

trel: oh, yes, that does appear to be.... exactly what we want already?

MartinFlores751:
Tested & Verified:
 - iget
 - imeta (pre & post change)
 - irmdir
 - irule
 - itree

Not sure how to get a pulse from ierror...

Side note, got irule to segfault.

src/imeta.cpp
MartinFlores751: The existing utils works just fine. Any preferences? If not, think we'll keep the old one.

alanking:
`ierror` doesn't connect to the server, so there will never be any entry in `ips` for `ierror` calls.

Please write up an issue for the `irule` segfault that you observed (or possibly see if it's the same as https://github.com/irods/irods/issues/7074)

src/imeta.cpp
alanking: Turns out, this was fixed in https://github.com/irods/irods/issues/7376 (specifically, https://github.com/irods/irods_client_icommands/pull/426). So we can either remove line 2142 in favor of your change, or just remove your change.

korydraughn: Agreed.

MartinFlores751: Removed new line.

MartinFlores751:
Think this is ready.

alanking (commented):
Excellent. Please squash commits, but omit # for now.

MartinFlores751:
Squashed.

korydraughn (approved):
Looks good to me.

Pound it.

alanking (approved):

MartinFlores751:
It's got the '#'.

alanking:
Please cherry-pick to 4-3-stable. Thanks

PR #477, opened 2024-05-10
[irods/irods#7592] Fix centos 7 build with new externals (main)

SwooshyCueb:
In service of irods/irods#7592

With the updated externals from irods/externals#246, building `ierror`, `ienv`, and `ipwd` fails on Centos 7 with the following error:
```
/opt/rh/devtoolset-10/root/usr/lib/gcc/x86_64-redhat-linux/10/../../../../bin/ld: /opt/rh/devtoolset-10/root/usr/lib/gcc/x86_64-redhat-linux/10/libstdc++_nonshared.a(system_error48.o): undefined reference to symbol '__cxa_free_exception@@CXXABI_1.3'
/opt/rh/devtoolset-10/root/usr/lib/gcc/x86_64-redhat-linux/10/../../../../bin/ld: /usr/lib64/libstdc++.so.6: error adding symbols: DSO missing from command line
clang-13: error: linker command failed with exit code 1 (use -v to see invocation)
```

I'm not sure what changed to cause this to happen, but passing `--no-as-needed` to the linker fixes it.

This PR does not need to wait on any other PRs before being merged.

PR #478, opened 2024-05-10
[irods/irods#7592] Fix centos 7 build against libstdc++ with new externals (4-3-stable)

SwooshyCueb:
4-3-stable version of #477

In service of irods/irods#7592

With the updated externals from irods/externals#246, building `ierror`, `ienv`, and `ipwd` against libstdc++ fails on Centos 7 with the following error:
```
/opt/rh/devtoolset-10/root/usr/lib/gcc/x86_64-redhat-linux/10/../../../../bin/ld: /opt/rh/devtoolset-10/root/usr/lib/gcc/x86_64-redhat-linux/10/libstdc++_nonshared.a(system_error48.o): undefined reference to symbol '__cxa_free_exception@@CXXABI_1.3'
/opt/rh/devtoolset-10/root/usr/lib/gcc/x86_64-redhat-linux/10/../../../../bin/ld: /usr/lib64/libstdc++.so.6: error adding symbols: DSO missing from command line
clang-13: error: linker command failed with exit code 1 (use -v to see invocation)
```

I'm not sure what changed to cause this to happen, but passing `--no-as-needed` to the linker fixes it.

This PR does not need to wait on any other PRs before being merged.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -287,6 +287,25 @@ foreach(EXECUTABLE ${IRODS_CLIENT_ICOMMANDS_EXECUTABLES})
   endif()
 endforeach()
 
+# Fix build on EL7
+if (IRODS_LINUX_DISTRIBUTION_NAME STREQUAL "centos" AND IRODS_LINUX_DISTRIBUTION_VERSION_MAJOR STREQUAL "7" AND NOT IRODS_BUILD_AGAINST_LIBCXX)
+  set(
+    ICOMMANDS_NEEDING_OVERLINK
+    ierror
+    ienv
+    ipwd
+  )
+  if (CMAKE_VERSION VERSION_LESS "3.13.0")
+    foreach(overlink_target IN LISTS ICOMMANDS_NEEDING_OVERLINK)
+      set_property(TARGET ${overlink_target} APPEND PROPERTY LINK_FLAGS "-Wl,--no-as-needed")
+    endforeach()
+  else()
+    foreach(overlink_target IN LISTS ICOMMANDS_NEEDING_OVERLINK)
+      target_link_options(${overlink_target} PRIVATE "LINKER:--no-as-needed")
+    endforeach()
+  endif()
+endif()
+
 set(
   IRODS_CLIENT_ICOMMANDS_SCRIPTS
   igetwild


alanking (approved):
Yep, looks identical.

CMakeLists.txt
```diff
@@ -287,6 +287,25 @@ foreach(EXECUTABLE ${IRODS_CLIENT_ICOMMANDS_EXECUTABLES})
   endif()
 endforeach()
 
+# Fix build on EL7
+if (IRODS_LINUX_DISTRIBUTION_NAME STREQUAL "centos" AND IRODS_LINUX_DISTRIBUTION_VERSION_MAJOR STREQUAL "7" AND NOT IRODS_BUILD_AGAINST_LIBCXX)
+  set(
+    ICOMMANDS_NEEDING_OVERLINK
```
korydraughn, line 293: What does this mean?

I want to make sure my mental model of `NEEDING_OVERLINK` is correct.

korydraughn (commented):
How did you land on `--no-as-needed` being the solution?

CMakeLists.txt
SwooshyCueb: it's just the name of the variable I'm using to store the list of icommands that need to be linked with `--no-as-needed`

SwooshyCueb:
> How did you land on `--no-as-needed` being the solution?

The symbol that the linker was complaining about was one from libstdc++, which the clang frontend puts toward the end of the arguments passed to the linker. I figured the linker was getting confused and tossing away some component of libstdc++ it's not supposed to, and `--no-as-needed` was the logical choice to make it not do that.

CMakeLists.txt
korydraughn: Sorry, I'm referring to the use of `OVERLINK`. What does that mean to you? I'm not familiar with that term.

korydraughn:
Got it. Thanks.

CMakeLists.txt
SwooshyCueb: Oh, it's not a standard term or anything. `--no-as-needed` tells the linker not to drop link libraries it deems unnecessary, so when it's used, it can result in the binary being linked to more libraries than it would have otherwise.

korydraughn: I see. Thanks for the explanation.

korydraughn (approved):
Pound it.

SwooshyCueb:
\#'d

PR #487, opened 2024-06-18
[irods/irods#7704] Use boost's trim, don't call `trimPrefix(...)` or `trimSpaces(...)` (main)

MartinFlores751:
This commit effectively "inlines" `trimPrefix(...)`, as the two locations in this commit are the only call sites for it. Additionally, replaces `trimSpaces(...)` with `trim_left_copy(...)` from boost.

### DIFF
--- a/src/irule.cpp
+++ b/src/irule.cpp
@@ -12,9 +12,12 @@
 #include <irods/irods_configuration_keywords.hpp>
 #include <irods/rcMisc.h>
 
+#include <boost/algorithm/string/trim.hpp>
 #include <boost/program_options.hpp>
 
 #include <cstdio>
+#include <cstring>
+#include <string>
 
 void usage();
 
@@ -319,14 +322,29 @@ main( int argc, char **argv ) {
                     continue;
                 }
 
-                if ( rulegen ) {
-                    if ( startsWith( buf, "INPUT" ) || startsWith( buf, "input" ) ) {
+                if (rulegen) {
+                    const auto skip_and_trim{[&buf](auto _skip_len) -> void {
+                        constexpr auto null_term{1};
+                        const auto copy_length{std::strlen(buf) + null_term - _skip_len};
+
+                        // Trim prefix
+                        std::memmove(buf, buf + _skip_len, copy_length);
+
+                        // Trim spaces
+                        auto res{boost::algorithm::trim_copy_if(
+                            std::string{buf}, [](unsigned char c) { return std::isspace(c); })};
+                        std::strncpy(buf, res.c_str(), res.size() + null_term);
+                    }};
+
+                    if (startsWith(buf, "INPUT") || startsWith(buf, "input")) {
                         gotRule = 1;
-                        trimSpaces( trimPrefix( buf ) );
+                        constexpr auto skip_len{5};
+                        skip_and_trim(skip_len);
                     }
                     else if ( startsWith( buf, "OUTPUT" ) || startsWith( buf, "output" ) ) {
                         gotRule = 2;
-                        trimSpaces( trimPrefix( buf ) );
+                        constexpr auto skip_len{6};
+                        skip_and_trim(skip_len);
                     }
                 }
 


src/irule.cpp
```diff
@@ -322,11 +325,25 @@ main( int argc, char **argv ) {
...
                         gotRule = 1;
-                        trimSpaces( trimPrefix( buf ) );
+                        constexpr auto skip_len{5};
+                        const auto copy_length{std::strlen(buf) + 1 - skip_len};
+
+                        // Blank out "INPUT"
+                        std::memmove(buf, buf + skip_len, copy_length);
+
+                        auto res{boost::algorithm::trim_left_copy(std::string{buf})};
+                        std::strncpy(buf, res.c_str(), res.size() + 1);
```
MartinFlores751, line 335: Consider adding `constexpr auto null_term{1};` to make the intent of the `+ 1` more obvious.

Possibly a better name for that too?

Same for other block.

MartinFlores751 (commented):
Note before I forget.

src/irule.cpp
```diff
@@ -322,11 +325,25 @@ main( int argc, char **argv ) {
                 if ( rulegen ) {
                     if ( startsWith( buf, "INPUT" ) || startsWith( buf, "input" ) ) {
                         gotRule = 1;
-                        trimSpaces( trimPrefix( buf ) );
+                        constexpr auto skip_len{5};
+                        const auto copy_length{std::strlen(buf) + 1 - skip_len};
```
korydraughn, line 329: Looks like this can be `constexpr` since `buf` is allocated on the stack.

src/irule.cpp
korydraughn, line 335: I'm a little confused by these lines. Please explain what's happening.

src/irule.cpp
```diff
@@ -322,11 +325,25 @@ main( int argc, char **argv ) {
...
+                        std::memmove(buf, buf + skip_len, copy_length);
+
+                        auto res{boost::algorithm::trim_left_copy(std::string{buf})};
+                        std::strncpy(buf, res.c_str(), res.size() + 1);
                     }
                     else if ( startsWith( buf, "OUTPUT" ) || startsWith( buf, "output" ) ) {
                         gotRule = 2;
-                        trimSpaces( trimPrefix( buf ) );
+                        constexpr auto skip_len{6};
+                        const auto copy_length{std::strlen(buf) + 1 - skip_len};
```
korydraughn, line 340: Can be `constexpr` too.

korydraughn (commented):
Looking good so far.

trel:
Also please check our other repositories for any usage… most specifically icommands.

src/irule.cpp
MartinFlores751: The top line:

It removes, in this case, "INPUT" from `buf` in place. You can see this also used in `trimPrefix(...)`.

We know it'll remove "INPUT" because of how `startsWith(...)` behaves:

https://github.com/irods/irods/blob/cf4744ecd539f3ea3847f957d8d1121a468100d9/lib/core/src/rcMisc.cpp#L4449-L4451

Relying on its behavior, we can "hard-code" the number of chars to skip (this is `skip_len`).

---

The bottom two lines:

`boost::algorithm::trim_left_copy(...)` is templated, so its input is its return type. This does not work with c-strings. So, `buf` gets converted to a `std::string`, is trimmed, and returned in `res`.

We then copy the trimmed `res` into `buf`, so that the trimmed result is available in the loop.

src/irule.cpp
MartinFlores751: I think since the string in `buf` is modified in-place, it cannot be `constexpr`.

korydraughn: Why does modification matter here? `buf` lives on the stack. The modification of the contents is independent of the storage.
```cpp
constexpr auto copy_length = sizeof(buf) - skip_len;
```

src/irule.cpp
korydraughn: Ok. According to cppreference, the call to `std::memmove` will essentially shift everything to the left by `skip_len` characters. That will result in `INPUT` being removed. That also means L334 and L335 aren't necessary.

See https://gcc.godbolt.org/z/xPv6bEhqd.

Am I missing a detail somewhere?

MartinFlores751: Yes, on `memmove`, it is for removing the matched string.

---

For missing details,

The original behavior was to both clear out the matched string, and clear out spaces afterwards. This behavior can be seen in both `trimPrefix(...)` and `trimSpaces(...)`, which were both originally called on `buf`. This should be visible in the removed code.

I don't know enough about rules to say if this is necessary, so replicating the original behavior was what I went for.

src/irule.cpp
MartinFlores751: I think I get you.

The original idea was to shift only the usable data in `buf`. It's why `copy_length` is not `sizeof(buf)`, and is instead `std::strlen(buf)`.

We can shift everything in `buf` to make it `constexpr`, which you're suggesting, correct?

src/irule.cpp
korydraughn: That's the correct approach.

As @trel has said, lean on the test suite. If things pass, then we can be confident we haven't gone backwards. The test suite allows us to experiment and has done a very good job at telling us when we've done too much.

src/irule.cpp
korydraughn: Hmm, let's keep what you have. I forgot that buffer is likely holding a string passed via stdin or a file.

alanking:
> Also please check our other repositories for any usage… most specifically icommands.

This is the iCommands repo, and this replaces the only 2 uses that I can see. But yes, maybe eyeball PREP and S3 just in case.

trel:
> This is the iCommands repo

well, when you put it that way...

alanking:
Eh, I went ahead and grepped for it in those two repos and didn't see any uses, so I think we are good on the main ones.

trel:
Excellent.   Commence getting that minus.

src/irule.cpp
korydraughn: Ignore this.

See conversation starting at https://github.com/irods/irods_client_icommands/pull/487#discussion_r1645185339.

src/irule.cpp
korydraughn: Marking as resolved.

src/irule.cpp
MartinFlores751: Done.

src/irule.cpp
```diff
@@ -320,13 +323,28 @@ main( int argc, char **argv ) {
...
                         gotRule = 1;
-                        trimSpaces( trimPrefix( buf ) );
+                        constexpr auto skip_len{5};
+                        const auto copy_length{std::strlen(buf) + null_term - skip_len};
+
+                        // Blank out "INPUT"
+                        std::memmove(buf, buf + skip_len, copy_length);
+
+                        auto res{boost::algorithm::trim_left_copy(std::string{buf})};
+                        std::strncpy(buf, res.c_str(), res.size() + null_term);
```
korydraughn, line 336: This looks identical to the body of the other if-stmt. Perhaps we can put it in a function?

src/irule.cpp
MartinFlores751: Resolving.

src/irule.cpp
```diff
@@ -13,8 +13,11 @@
 #include <irods/rcMisc.h>
 
 #include <boost/program_options.hpp>
+#include <boost/algorithm/string/trim.hpp>
```
alanking, line 16: Alphabetize, please

src/irule.cpp
alanking: Looks like this has been addressed

alanking (commented):
No real concerns from me.

src/irule.cpp
MartinFlores751: Resolving.

src/irule.cpp
MartinFlores751: Done.

korydraughn:
Are the tests passing? Is this ready for squashing?

MartinFlores751:
Last tested commit was 'correct trimming behavior', the rest of the commits following are refactors. I could retest using tests that failed when fixing things (`test_all_rules` and `test_irule`) to see if I made a mistake when refactoring.

korydraughn:
I see. If you're confident things are correct, squash it. No #'s.

If there's an issue, we'll catch it during release testing.

MartinFlores751:
Just going to rescope one variable, then squash.

alanking (commented):
Looks good... Please squash at your ready. Omit #s

MartinFlores751:
Squashed.

alanking (approved):
Looks good to me... # it

MartinFlores751:
Pounded.

korydraughn:
Let's get that cherry-pick 4-3-stable going.

PR #488, opened 2024-06-19
[#7761] Fix help text for itree JSON option (4-3-stable)

korydraughn:
This doesn't require testing. It's just help text.

I opened an issue for test coverage. See irods/irods#7825. We'll handle that post 4.3.3.

### DIFF
--- a/src/itree.cpp
+++ b/src/itree.cpp
@@ -361,7 +361,7 @@ Options:
       --indent=INTEGER
                   The number of spaces used for indenting nested collections.
                   (defaults to 2).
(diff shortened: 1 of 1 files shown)


alanking (approved):
Well, it requires test insofar as you've confirmed that `-J` performs the function that `-j` was previously thought to have been performing, right? If so, let's put this in.

alanking:
Okay. I say we add #s.

korydraughn:
#'d

trel:
ew.  don't like that it's a capital.

but here we are.

PR #494, opened 2024-07-10
[irods/irods#7759] Do not stamp iRODS version number on icommands package (main)

SwooshyCueb:
Companion to irods/irods#7884
In service of irods/irods#7759


SwooshyCueb:
\#'d

PR #496, opened 2024-07-15
[irods/irods_7724] add filesize clarification for ichksum behavior (4-3-stable)

trel:


trel:
hmm.   well, i think we cannot read past the end of the file, regardless.

alanking:
Okay. I think my brain wanted to actually ask if we should include a disclaimer about checksums still appearing to be correct even if the size in the catalog is incorrect, which could probably be classified as some sort of data "integrity" issue. You get an error (or whatever is happening in irods/irods#7724) when the size in the catalog is smaller than in reality, but you get no error or indication that anything is wrong when the size in the catalog is larger than in reality. I guess that's more of a job for `ifsck` than `ichksum`.

trel:
> I guess that's more of a job for `ifsck` than `ichksum`.

I think this is correct.   Want to keep this one narrow for now.

alanking (approved):
Seems good to me

trel:
#ing nowish

trel:
#'d

korydraughn (approved):

PR #502, opened 2024-09-24
[irods/irods_2600] update irsync help text

trel:
just some words.


src/irsync.cpp
```diff
@@ -169,53 +169,53 @@ usage() {
...
         "     cp -r foo1 foo2",
         " ",
-        "has a different meaning depending on whether the target directory foo2 ",
-        "exists. If the target directory exists, the content of source directory foo1",
-        "is copied to the target directory foo2/foo1. But if the target directory",
-        "does not exist, the content is copied to the target directory foo2.",
+        "has a different meaning depending on whether the target directory foo2",
+        "already exists. If the target directory exists, the content of source directory",
+        "foo1 is copied into the target directory foo2/foo1. But if the target directory",
+        "foo1 does not already exist, the content is copied as the target directory foo2.",
```
korydraughn, line 193: `foo1` is identified as source directory and target directory across these four lines.

Seems it should only be referred to as the source directory?

korydraughn (commented):
Changes look good overall.

src/irsync.cpp
trel: oh... i think there's a bug in the words.   looking again...

trel: yep, bug found - good eye.   updated in second commit.

src/irsync.cpp
```diff
@@ -144,12 +144,12 @@ usage() {
         "          sourceFile|sourceDirectory [....] targetFile|targetDirectory",
         " ",
         "Synchronize the data between a local copy (local file system) and",
-        "the copy stored in iRODS or between two iRODS copies. The command can be ",
-        "in one of the three modes  : synchronization of data from the client's",
+        "the copy stored in iRODS or between two iRODS copies. The command can be",
+        "in one of the three modes : synchronization of data from the client's",
```
alanking, line 148:
```suggestion
        "in one of the three modes: synchronization of data from the client's",
```

alanking (commented):
My suggestion is quite ignore-able

korydraughn (approved):
Looks good to me.

I think we're ready for the pound.

trel:
#'d

PR #512, opened 2024-12-13
[irods/irods7961] Deprecate imeta interactive mode (4-3-stable)

FifthPotato:
Couldn't find reference to it in many other places. Wondering if it's reasonable to print the warning message every time interactive mode is used.

### DIFF
--- a/src/imeta.cpp
+++ b/src/imeta.cpp
@@ -2204,6 +2204,7 @@ int main( int argc, const char **argv )
     }
 
     if ( command_mode ) {
+        printf("NOTE: Interactive mode is deprecated and will be removed in an upcoming release.\n");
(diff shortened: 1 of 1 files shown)


trel:
> Wondering if it's reasonable to print the warning message every time interactive mode is used.

I think so.   The diff you have currently will only print it when first entering the mode (when no subcommand was issued).

I think this is sufficient.

alanking (commented):
Looks good. Squash?

FifthPotato:
Squished!

korydraughn (approved):
Pound it.

alanking (approved):

FifthPotato:
Octothorpe'd

PR #516, opened 2025-01-28
[irods/irods8120] Remove duplicated rcDisconnect call (main)

FifthPotato:
Seems to be causing the segfault. Will run tests to make sure this doesn't break anything.

### DIFF
--- a/src/iqmod.cpp
+++ b/src/iqmod.cpp
@@ -112,7 +112,6 @@ main( int argc, char **argv ) {
                              argv[argOffset + 2] );
 
     printErrorStack( Conn->rError );
-    rcDisconnect( Conn );
 
     if ( status != 0 ) {
         return 4;


trel:
Eyeballing the other edits in https://github.com/irods/irods_client_icommands/commit/88fd0a9f2afd4b538fa7bfa115afa1edf0cd9beb for https://github.com/irods/irods/issues/6390 -- it seems this might be the only one that was missed.

Please confirm.

FifthPotato:
> Is it generally true that calling `rcDisconnect` multiple times causes a segfault in the client? If so, can we see why and prevent it?

Not sure about the "generally true" bit, but I can report what `gdb` gave me. At the end of `rcDisconnect`, `freeRcComm` is called, which calls `cleanRcComm` which sets `thread_ctx` to NULL.
https://github.com/irods/irods/blob/cdd3c3937c23bec6a5cb322c963def5a990d05fe/lib/core/src/rcConnect.cpp#L351C5-L351C29
When it's called again, https://github.com/irods/irods/blob/cdd3c3937c23bec6a5cb322c963def5a990d05fe/lib/core/src/rcConnect.cpp#L289 segfaults.

FifthPotato:
> Eyeballing the other edits in [88fd0a9](https://github.com/irods/irods_client_icommands/commit/88fd0a9f2afd4b538fa7bfa115afa1edf0cd9beb) for [irods/irods#6390](https://github.com/irods/irods/issues/6390) -- it seems this might be the only one that was missed.
> 
> Please confirm.

It would seem like it. Quick eyeball seems to imply that too.

alanking:
Nice detective work. I think we should go ahead and fix that possibility in `rcDisconnect` in addition to what you have here. Perhaps a simple pointer check before the dereference will be enough, i.e.:
```diff
-    if ( _conn->thread_ctx->reconnThr ) {
+    if (_conn->thread_ctx && _conn->thread_ctx->reconnThr) {
```

alanking (approved):
I think this change is good to go. # it

FifthPotato:
Octothorpe'd!

PR #532, opened 2025-02-11
[irods/irods#2978] iadmin: Add support to "add" a "group" (main)

alanking:
In service of irods/irods#2978

Companion PR: https://github.com/irods/irods/pull/8192

Leaving in draft til companion PR is opened.

### DIFF
--- a/src/iadmin.cpp
+++ b/src/iadmin.cpp
@@ -1620,8 +1620,15 @@ doCommand( char *cmdToken[], rodsArguments_t* _rodsArgs = 0 ) {
     }
 
     if ( strcmp( cmdToken[0], "mkgroup" ) == 0 ) {
-        generalAdmin( 0, "add", "user", cmdToken[1], "rodsgroup",
-                      "", "", "", "", "", "" );
+        const auto version = irods::to_version(Conn->svrVersion->relVersion);
+        if (version && version.value() > irods::version{4, 3, 3}) {
+            // If we are adding a group, we don't need to specify a user type - it's a group.
+            generalAdmin(0, "add", "group", cmdToken[1], "", "", "", "", "", "", "");
+        }
+        else {
+            // Versions 4.3.3 and earlier only know how to "add" a "user" with a "rodsgroup" type.
+            generalAdmin(0, "add", "user", cmdToken[1], "rodsgroup", "", "", "", "", "", "");
+        }
         return 0;
     }
     if ( strcmp( cmdToken[0], "rmgroup" ) == 0 ) {


src/iadmin.cpp
```diff
@@ -1620,8 +1620,22 @@ doCommand( char *cmdToken[], rodsArguments_t* _rodsArgs = 0 ) {
     }
 
     if ( strcmp( cmdToken[0], "mkgroup" ) == 0 ) {
-        generalAdmin( 0, "add", "user", cmdToken[1], "rodsgroup",
-                      "", "", "", "", "", "" );
+        const auto version = irods::to_version(Conn->svrVersion->relVersion);
+        const bool using_add_group = (version && version.value() > irods::version{4, 3, 3});
```
korydraughn, line 1624: Consider replacing `using` with `use`.

src/iadmin.cpp
```diff
@@ -1620,8 +1620,22 @@ doCommand( char *cmdToken[], rodsArguments_t* _rodsArgs = 0 ) {
...
+        const char* entity_type = using_add_group ? "group" : "user";
+        // If we are adding a group, we don't need to specify a user type.
+        const char* user_type = using_add_group ? "" : "rodsgroup";
```
korydraughn, line 1627: Consider grouping these together via an if-else block to make it even clearer to readers that these are related.

src/iadmin.cpp
```diff
@@ -1620,8 +1620,22 @@ doCommand( char *cmdToken[], rodsArguments_t* _rodsArgs = 0 ) {
...
+        generalAdmin(0,
+                     "add",
+                     const_cast<char*>(entity_type),
+                     cmdToken[1],
+                     const_cast<char*>(user_type),
```
korydraughn, line 1632: Could use `std::string::data` to remove the need for `const_cast`.

src/iadmin.cpp
alanking: I think using an if-else block addressed all three comments. Please let me know if you had something else in mind.

src/iadmin.cpp
korydraughn: See https://github.com/irods/irods_client_icommands/pull/532#discussion_r1951291560

src/iadmin.cpp
korydraughn: See https://github.com/irods/irods_client_icommands/pull/532#discussion_r1951291560

korydraughn (approved):
Looks correct and good to me.

I say pound it when the companion PR is approved.

alanking:
#'d, mergin

PR #547, opened 2025-03-18
[irods/irods#8260] Allow modification of access time via iadmin modrepl (main)

korydraughn:
Depends on irods/irods#8269.


alanking (approved):
Looks good to me. # when ready. Will merge along with https://github.com/irods/irods/pull/8269.

PR #557, opened 2025-04-10
Do not set client-server negotiation policy / Guarantee rodsEnv strings are null-terminated (main)

korydraughn:
Depends on irods/irods#8357.

Will test soon.


alanking (commented):
Looks good to me. Let's get an issue number on that commit and then I think this is good to just go in?

alanking (approved):
\# it

korydraughn:
Pounded.

PR #558, opened 2025-04-14
[irods/irods#8379] Guarantee rodsEnv strings are null-terminated (4-3-stable)

korydraughn:
Cherry-picked from #557.

Had to handle one difference between branches, but otherwise, I think this is ready for squashing. Updated all calls involving `strncpy`.

I think the changes are obvious and do not require additional testing.


alanking (approved):
\# it

korydraughn:
Pounded.

PR #565, opened 2025-05-01
[irods/irods#8334] iput: Fix leak reported by ASAN (main)

alanking:
In service of irods/irods#8334

Companion PR: (irods/irods, coming soon)

Still need to run tests, so putting in draft

### DIFF
--- a/src/iput.cpp
+++ b/src/iput.cpp
@@ -32,6 +32,7 @@ main( int argc, char **argv ) {
     }
 
     rodsPathInp_t rodsPathInp{};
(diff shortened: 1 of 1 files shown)


alanking:
This change is causing a double-free, so it will require a deeper investigation to fix this leak. Seeing as the leak reported from `iput` is constant (regardless of `iput -r` or just `iput`), I think we should close this or let it sit for now. We can implement a proper fix later.

trel:
draft... and let it sit seems fine.

PR #568, opened 2025-05-08
[irods/irods8487] iinit to print connection information (4-3-stable)

trel:
tested at the bench... no test suite run.


trel:
oh, scratch this - we said 5.0.0 only...

PR #569, opened 2025-05-08
[irods/irods8487] iinit to print connection information (main)

trel:
cherry-pick of https://github.com/irods/irods_client_icommands/pull/568


korydraughn (commented):
Change looks good to me.

@alanking Do you sense this will play well with output from the auth plugins?

trel:
I placed the printf 'higher' than before... just after the environment was reloaded... so it should now appear before any prompts / cycles with the server.

alanking:
> @alanking Do you sense this will play well with output from the auth plugins?

I think it should be fine. Like @trel said, the output will appear only after the client environment is established but before prompting for any authentication information. So, if the user (like the use case which inspired this) already has a few environments being swapped out, it will just show which user/server is being authenticated and then the password prompt will be shown (at least for native/pam_password).

korydraughn:
Yep, sounds good.

korydraughn (approved):
I think we're good to pound it if @alanking agrees.

alanking (approved):
\# it

trel:
#'d

PR #591, opened 2025-09-17
[irods/irods#8653] imiscsvrinfo: Report error when connected to 4.3.3 and earlier (5-0-stable)

korydraughn:
Will test soon.


korydraughn:
Confirmed this works as intended.

The only decision to be made now is whether we like the message.

trel:
message is clear - seems good enough.

korydraughn:
Okay. Pounding and merging.

PR #598, opened 2026-01-08
[irods/irods#8671] Return nonzero when get/set_grid_configuration APIs report an error (4-3-stable)

korydraughn:
Companion PR: https://github.com/irods/irods/pull/8792


alanking (commented):
This part looks good. Will approve along with the companion PR.

PR #604, opened 2026-02-04
[irods/irods#8618] Update help text for physical quota commands (4-3-stable)

korydraughn:


src/iadmin.cpp
```diff
@@ -2604,22 +2604,39 @@ usage( char *subOpt ) {
         "The Value parameter is the quota amount in bytes.",
         "Set this to 0 to remove a group quota.",
         " ",
+        "Administrators are encouraged to run 'iadmin cu' after setting a",
```
alanking: I feel like this wording could be stronger...

"Administrators should run 'iadmin cu' after setting a..."

Thoughts?

trel: seems better, yes.

korydraughn: Done.

alanking (commented):
Looks good. Squash it

korydraughn:
Squashed.

alanking (approved):
\# it

korydraughn:
Added pound.

PR #606, opened 2026-02-13
[irods/irods#8850] Deprecate iadmin subcommands: ctime, mkdir, rmdir, ls (4-3-stable)

korydraughn:


FifthPotato (approved):
String additions look fine to me.

korydraughn:
None of our other icommands specify in which version of iRODS a certain thing was deprecated. I'm going to tweak the message to say the following:

```
This command is deprecated.
```

korydraughn:
I also added a note to the top of the iadmin help text and added some empty lines to separate sections.

src/iadmin.cpp
```diff
@@ -2287,6 +2296,9 @@ usage( char *subOpt ) {
 
     char *mkdirMsgs[] = {
         "***************************** WARNING ********************************",
+        " ",
+        "This command is deprecated.",
```
FifthPotato, line 2298: Do we want this message to be above or below the large `****WARNING****` above? I think the warning applies more strongly to the fact that the command "should never be called directly by a user." What do you think?

FifthPotato (approved):
String changes still look good. One optional thing to look at.

src/iadmin.cpp
korydraughn: That's true. Maybe we put it above the warning message.

Thoughts?

korydraughn: At least then, the warning is tied to the paragraph like in the original text.

FifthPotato: Yeah, I was thinking above. If we're in agreement, that seems like a good change.

korydraughn: Will move it.

korydraughn: Done.

korydraughn:
@FifthPotato Please see latest change to make sure it's good.

FifthPotato:
> @FifthPotato Please see latest change to make sure it's good.

Looks good to me!

korydraughn:
Confirmed that the help text for each command shows the expected text.

Squashed and pounded.

Merging.

PR #608, opened 2026-02-26
[irods/irods#8873] Declare dependency on findutils on EL (4-3-stable)

SwooshyCueb:
Addresses irods/irods#8873


korydraughn (commented):
Does that change affect EL9's dependencies?

SwooshyCueb:
> Does that change affect EL9's dependencies?

Yes.

korydraughn:
In your opinion, how much should we care about that affecting EL9?

How much work does it take to limit it to EL8? That would make it extremely clear which OS that change applies to.

SwooshyCueb:
I think it should apply to both EL8 and EL9, as `xargs` is provided by this package on both versions.

korydraughn:
So even though the tests for EL9 pass, it's more of a future proofing play for EL9.

And given this is about packaging, it's relatively low risk.

Does that align with your reasoning/approach?

SwooshyCueb:
Almost. I wouldn't say future proofing. It's more that: just because findutils is installed by default in the Rocky 9 docker images doesn't mean it's guaranteed to be on a user's system.

alanking (approved):
Seems good to me. Deferring to @korydraughn

korydraughn:
Ohhh, I see now.

I agree with your assessment.

korydraughn (approved):
Pound it.

SwooshyCueb:
\#'d

PR #609, opened 2026-04-28
[irods/irods7943] Added the -j option for itree.

JustinKyleJames:
The -J option is being deprecated.  This information has been added to the itree -h text.


korydraughn:
The icommands source is now part of irods/irods.

You need to apply these changes to  https://github.com/irods/irods/blob/main/clients/icommands/src/itree.cpp instead. Then you can close this PR.

JustinKyleJames:
> these

Forgot that.  I copied it over.

